PR #22, opened 2019-04-09
show color handles

goswami-rahul:
- [x] separate prettyhandles from showhandles


poetry.lock
```diff
@@ -279,6 +279,14 @@ optional = false
...
 version = "1.3.0"
 
+# [[package]]
+# category = "main"
+# description = "Image library"
+# name = "pillow"
+# optional = false
+# python-versions = ">=3.6"
+# version = "5.4.1"
+
```
goswami-rahul: @meooow25 help me out with this. Is it fine to uncomment these? How do I add the hash?

meooow25: The `poetry.lock` file is autogenerated by poetry. If you want to add a dependency it should be in the `pyproject.toml` file.

PR #30, opened 2019-04-27
plot cf rating distribution

krofna:


krofna:
looks like changing figsize affects other plots. how to fix that?

tle/cogs/graphs.py
```diff
@@ -362,6 +362,52 @@ def __init__(self, bot):
         await cf_common.cf_handle_error_handler(ctx, error)
         await cf_common.run_handle_coro_error_handler(ctx, error)
 
+    @plot.command(brief='Show codeforces rating distribution')
```
meooow25: Keep the command above the `cog_command_error` function.

tle/cogs/graphs.py
```diff
@@ -362,6 +362,52 @@ def __init__(self, bot):
...
+    async def cfdistrib(self, ctx, mode: str = 'log'):
+        params = {'activeOnly' : 'true'}
```
meooow25: Unused, remove.

tle/cogs/graphs.py
```diff
@@ -362,6 +362,52 @@ def __init__(self, bot):
...
+        resp = await cf_common.cache.get_user_rating(3600)
+        ratings = [r if r >= 0 else 0 for r in resp.values()]
```
meooow25: Use `max(r, 0)` instead.

tle/cogs/graphs.py
```diff
@@ -362,6 +362,52 @@ def __init__(self, bot):
...
+        cent.reverse()
+
+        label = []
+        for r, c in zip(x, cent):
+            label.append('{} ({})'.format(r, c))
+
+        plt.clf()
+        if mode == 'log':
+            plt.figure(figsize=(15, 5))
+        else:
```
meooow25: I think it's better to have two fixed states, either `'log'` or, say, `'normal'` instead of just anything else meaning normal.

tle/cogs/graphs.py
```diff
@@ -362,6 +362,52 @@ def __init__(self, bot):
         await cf_common.cf_handle_error_handler(ctx, error)
         await cf_common.run_handle_coro_error_handler(ctx, error)
 
+    @plot.command(brief='Show codeforces rating distribution')
+    async def cfdistrib(self, ctx, mode: str = 'log'):
+        params = {'activeOnly' : 'true'}
+        resp = await cf_common.cache.get_user_rating(3600)
+        ratings = [r if r >= 0 else 0 for r in resp.values()]
+        x = [k * 100 for k in range(39)]
```
meooow25: It would be nice to declare `bins = 39` and use that below wherever required.

tle/cogs/graphs.py
```diff
@@ -362,6 +362,52 @@ def __init__(self, bot):
...
+        colors = []
+        rnum = [12, 2, 2, 3, 2, 2, 1, 14]
```
meooow25: This seems very cryptic and will break if ranks are modified later. I suggest replacing it with

```py
low, high = 0, 100 * bins
for rank in cf.RATED_RANKS:
    for r in range(max(rank.low, low), min(rank.high, high), 100):
        colors.append('#' + '%06x' % rank.color_embed)
assert len(colors) == bins, f'Expected {bins} colors, got {len(colors)}'
```

tle/cogs/graphs.py
```diff
@@ -362,6 +362,52 @@ def __init__(self, bot):
...
+        for rank, n in zip(cf.RATED_RANKS, rnum):
+            colors += ['#' + '%06x' % rank.color_embed] * n
+
+        height = [0 for _ in range(39)]
```
meooow25: `height = [0] * 39`

tle/cogs/graphs.py
```diff
@@ -362,6 +362,52 @@ def __init__(self, bot):
...
+            cent.append(round(100 * (1 - csum / users)));
+        cent.reverse()
+
+        label = []
+        for r, c in zip(x, cent):
+            label.append('{} ({})'.format(r, c))
+
+        plt.clf()
+        if mode == 'log':
+            plt.figure(figsize=(15, 5))
```
meooow25: To prevent the figsize change from affecting other plots, keep the figure created as `fig` for example, and `plt.close(fig)` after the image is ready.

tle/cogs/graphs.py
```diff
@@ -358,6 +358,55 @@ def __init__(self, bot):
...
+        assert len(colors) == bins, f'Expected {bins} colors, got {len(colors)}'
+
+        height = [0] * bins
+        for r in ratings:
+            height[r // 100] += 1
+
+        csum = 0
+        cent = []
+        users = sum(height)
+        for k in reversed(range(bins)):
```
algmyr: loop over iterables and not indices where feasible `for h in reversed(height):` or `for h in height[::-1]:`

tle/cogs/graphs.py
```diff
@@ -358,6 +358,55 @@ def __init__(self, bot):
...
+        label = []
+        for r, c in zip(x, cent):
+            label.append('{} ({})'.format(r, c))
+
+        plt.clf()
+        if mode == 'log':
+            fig = plt.figure(figsize=(15, 5))
+        elif mode == 'normal':
+            fig = plt.figure(figsize=(20,100))
+            plt.locator_params(axis='y', nbins=50)
```
algmyr: What happens if mode is neither? (Bad things.)
Add an else and send an error to the user and return.

tle/cogs/graphs.py
```diff
@@ -358,6 +358,55 @@ def __init__(self, bot):
...
+        cent = []
+        users = sum(height)
+        for k in reversed(range(bins)):
+            csum += height[k]
+            cent.append(round(100 * (1 - csum / users)));
+        cent.reverse()
+
+        label = []
+        for r, c in zip(x, cent):
+            label.append('{} ({})'.format(r, c))
```
algmyr: Use a list comprehension and f-strings?
`label = [f'{r} ({c})' for r,c in zip(x, cent)]`

tle/cogs/graphs.py
```diff
@@ -358,6 +358,55 @@ def __init__(self, bot):
...
+    async def cfdistrib(self, ctx, mode: str = 'log'):
+        resp = await cf_common.cache.get_user_rating(3600)
+        ratings = [max(r, 0) for r in resp.values()]
+        bins = 39
+        x = [k * 100 for k in range(bins)]
+        colors = []
+        low, high = 0, 100 * bins
+        for rank in cf.RATED_RANKS:
+            for r in range(max(rank.low, low), min(rank.high, high), 100):
+                colors.append('#' + '%06x' % rank.color_embed)
```
algmyr: Why are we using `color_embed` rather than `color_graph` here?
```
colors.append(rank.color_graph)
```

tle/cogs/graphs.py
algmyr: I'm missing the forest for the trees, why not just do
```
for h in height:
    csum += h
    cent.append(round(100 * csum / users))
```

tle/cogs/graphs.py
algmyr: The fix to this broke stuff, see my comment later on

PR #35, opened 2019-05-01
Refactor cfdistrib and distrib to use same improved code

algmyr:
This adds color to `distrib` and uses the same code for `distrib` and `cfdistrib`. This also changes `mode` to just affect the y-scaling and not the figure size.

### DIFF
--- a/tle/cogs/graphs.py
+++ b/tle/cogs/graphs.py
@@ -232,23 +232,84 @@ class Graphs(commands.Cog):
         discord_common.set_author_footer(embed, ctx.author)
         await ctx.send(embed=embed, file=discord_file)
 
-    @plot.command(brief='Show server rating distribution')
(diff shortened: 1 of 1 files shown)


meooow25:
I think we can combine them as a command group. They they can be run as `distrib cf` and `distrib server`.

tle/cogs/graphs.py
```diff
@@ -232,23 +232,84 @@ def __init__(self, bot):
...
+        height = [0] * bins
+        for r in ratings:
+            height[r // binsize] += 1
+
+        csum = 0
+        cent = []
+        users = sum(height)
+        for h in height:
+            csum += h
+            cent.append(round(100 * csum / users))
```
krofna: division by zero if there are no users, but we probably don't care about that

Issue #60, opened 2019-06-22
Group "handle" commands


meooow25:
Group "handle" commands `sethandle`, `gethandle`, `showhandles`, `removehandle` and `prettyhandles` together.

PR #64, opened 2019-06-22
Added set, remove, get, show, pretty in a 'handle' group

cfalas:
This solves issue #60

### DIFF
--- a/tle/cogs/handles.py
+++ b/tle/cogs/handles.py
@@ -209,7 +209,7 @@ class Handles(commands.Cog):
         await ctx.send(msg)
 
     @handle.command(brief="Show all handles")
-    async def show(self, ctx):
+    async def list(self, ctx):
         """Shows all members of the server who have registered their handles and
         their Codeforces ratings.
         """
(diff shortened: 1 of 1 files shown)


krofna:
@meooow25 should `_updateroles` even be in this group?

cfalas:
I assumed that because it has to do with roles, therefore handles it should be here, but it's not that related so idk

cfalas:
Made the changes mentioned (removed _updateroles from the group and renamed show to list)

tle/cogs/handles.py
```diff
@@ -124,9 +124,13 @@ class Handles(commands.Cog):
     def __init__(self, bot):
         self.bot = bot
 
-    @commands.command(brief='sethandle [name] [handle] (admin-only)')
+    @commands.group(brief='Commands that have to do with handles', invoke_without_command=True)
+    async def handle(seld, ctx):
```
meooow25: Small typo, `self`.
Also, if you add `await ctx.send_help(ctx.command)` to the body, the command group will send its help message when invoked without subcommands.

tle/cogs/handles.py
```diff
@@ -124,9 +124,13 @@ class Handles(commands.Cog):
...
+        """Change or collect information about specific handles on Codeforces"""
+
+    @handle.command(brief='set [name] [handle] (admin-only)')
```
meooow25: Would be nice to update all the `brief`s since you're working in this area.

PR #66, opened 2019-06-23
Add a basic starboard

meooow25:
## A simple starboard cog :star: 

A starboard cog with the following functionality
- Set a particular channel as starboard.
- Reacting to a message with a :star: will post the message to a starboard channel if the reactor(?) is an admin or if the message now has >= 3 stars.
- Reacting to the same message after that has no effect on the starboard.
- Deleting the message from the starboard will allow the original message to be posted again when reacted to.

### Changes
- Added two tables, one for starboard settings and one the starboard messages.
- Added `Starboard` cog.

### Screenshot
screenshot

### DIFF
--- a/tle/cogs/starboard.py
+++ b/tle/cogs/starboard.py
@@ -1,0 +1,135 @@
+import asyncio
+import logging
+
+import discord
+from discord.ext import commands
+
+from tle.util import codeforces_common as cf_common
+from tle.util import discord_common
+
+_STAR = '\N{WHITE MEDIUM STAR}'
+_STAR_ORANGE = 0xffaa10
+_STAR_THRESHOLD = 3
+
+
+class StarboardCogError(commands.CommandError):
+    pass
+
+
+class Starboard(commands.Cog):
+    def __init__(self, bot):
+        self.bot = bot
+        self.locks = {}
+        self.logger = logging.getLogger(self.__class__.__name__)
+
+    @commands.Cog.listener()
+    async def on_raw_reaction_add(self, payload):
+        if str(payload.emoji) != _STAR or payload.guild_id is None:
+            return
+        res = cf_common.user_db.get_starboard_settings(payload.guild_id)
+        if res is None:
+            return
+        starboard_channel_id = int(res[0])
+        try:
+            await self.check_and_add_to_starboard(starboard_channel_id, payload)
+        except StarboardCogError as e:
+            self.logger.info(f'Failed to starboard: {e!r}')
+
+    @commands.Cog.listener()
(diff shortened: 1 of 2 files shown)


algmyr:
If it's possible, maybe move the content to right below the image so that things look more like the original post?

tle/cogs/starboard.py
```diff
@@ -0,0 +1,135 @@
...
+        if payload.channel_id != starboard_channel_id:
+            return
+        cf_common.user_db.remove_starboard_message(payload.message_id)
+        self.logger.info(f'Removed message {payload.message_id} from starboard')
+
+    @staticmethod
+    def prepare_embed(message):
+        # Adapted from https://github.com/Rapptz/RoboDanny/blob/rewrite/cogs/stars.py
+        embed = discord.Embed(color=_STAR_ORANGE, timestamp=message.created_at)
+        if message.content:
```
algmyr: As commented, would it look nice if this was below channel and link to original?

tle/cogs/starboard.py
```diff
@@ -0,0 +1,135 @@
...
+                    invoke_without_command=True)
+    async def starboard(self, ctx):
+        """Group for commands involving the starboard"""
+        await ctx.send_help(ctx.command)
+
+    @starboard.command(brief='Set starboard to current channel')
+    @commands.has_role('Admin')
+    async def here(self, ctx):
+        """Set the current channel as starboard"""
+        cf_common.user_db.set_starboard_settings(ctx.guild.id, ctx.channel.id)
```
algmyr: Can this add multiple starboards? If so, is this intended or will it lead to weird behavior? If not, maybe add a check and show an error.

tle/cogs/starboard.py
```diff
@@ -0,0 +1,135 @@
...
+            embed.add_field(name='Content', value=message.content, inline=False)
+
+        if message.embeds:
+            data = message.embeds[0]
+            if data.type == 'image':
+                embed.set_image(url=data.url)
+
+        if message.attachments:
+            file = message.attachments[0]
+            if file.url.lower().endswith(('png', 'jpeg', 'jpg', 'gif', 'webp')):
```
algmyr: could add webm for completeness

meooow25: Sadly, Discord can embed videos but we cannot :(

tle/cogs/starboard.py
meooow25: Agreed, and done.

tle/cogs/starboard.py
meooow25: Discussed and resolved.

PR #71, opened 2019-07-06
Various improvements to Handles cog.

meooow25:
Improvements to Handles cog

- Cog error system using `HandleCogError`.
- Truncate long usernames in `handle list`.
- Setting handle now also assigns the appropriate role.
- Removed all bare excepts.
- It is no longer necessary to have all CF rank roles for the cog to work.
- Other minor changes.

### DIFF
--- a/tle/cogs/handles.py
+++ b/tle/cogs/handles.py
@@ -1,4 +1,3 @@
-import logging
 import io
 
 import discord
@@ -13,7 +12,12 @@ from tle.util import table
 from PIL import Image, ImageFont, ImageDraw
 
 _HANDLES_PER_PAGE = 15
+_NAME_MAX_LEN = 20
 _PAGINATE_WAIT_TIME = 5 * 60  # 5 minutes
+
+
+class HandleCogError(commands.CommandError):
+    pass
 
 
 def rating_to_color(rating):
@@ -81,12 +85,11 @@ def get_prettyhandles_image(rankings):
 
 
 def _make_profile_embed(member, user, *, mode):
+    assert mode in ('set', 'get')
     if mode == 'set':
-        desc = f'Handle for **{member.display_name}** successfully set to **[{user.handle}]({user.url})**'
-    elif mode == 'get':
-        desc = f'Handle for **{member.display_name}** is currently set to **[{user.handle}]({user.url})**'
+        desc = f'Handle for {member.mention} successfully set to **[{user.handle}]({user.url})**'
     else:
-        return None
+        desc = f'Handle for {member.mention} is currently set to **[{user.handle}]({user.url})**'
     if user.rating is None:
         embed = discord.Embed(description=desc)
         embed.add_field(name='Rating', value='Unrated', inline=True)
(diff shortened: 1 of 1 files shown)


tle/cogs/handles.py
```diff
@@ -109,10 +112,12 @@ def _make_pages(users):
         t += table.Header('#', 'Name', 'Handle', 'Rating')
         t += table.Line()
         for i, (member, handle, rating) in enumerate(chunk):
+            name = member.display_name
+            if len(name) > _NAME_MAX_LEN:
+                name = name[:_NAME_MAX_LEN - 3] + '...'
```
algmyr: Consider using an ellipsis character `…` and ` - 1` for maximum space savings!

tle/cogs/handles.py
```diff
@@ -129,85 +134,84 @@ def __init__(self, bot):
         """Change or collect information about specific handles on Codeforces"""
         await ctx.send_help(ctx.command)
 
-    @handle.command(brief='Set Codeforces handle of a user (admin-only)')
+    async def update_member_rank_role(self, member, role_to_assign):
+        has_role = False
+        role_names_to_remove = set(rank.title for rank in cf.RATED_RANKS)
```
algmyr: Use {} instead of set() for comprehensions

tle/cogs/handles.py
```diff
@@ -129,85 +134,84 @@ def __init__(self, bot):
...
+        to_remove = []
+        for role in member.roles:
+            if role == role_to_assign:
+                has_role = True
+            elif role.name in role_names_to_remove:
+                to_remove.append(role)
```
algmyr: At the cost of two passes, maybe write the above with comprehensions and similar? Feels more readable
```
role_names_to_remove = {rank.title for rank in cf.RATED_RANKS} - {role_to_assign}
has_role = role_to_assign in member.roles
to_remove = [role for role in member.roles if role in role_names_to_remove]
```

tle/cogs/handles.py
```diff
@@ -129,85 +134,84 @@ def __init__(self, bot):
...
-
         embed = _make_profile_embed(member, user, mode='set')
         await ctx.send(embed=embed)
 
+        if user.rank == cf.UNRATED_RANK:
+            return
+        roles = [role for role in ctx.guild.roles if role.name == user.rank.title]
+        if not roles:
+            raise HandleCogError(f'Role for rank `{user.rank.title}` not present in the server')
+        await self.update_member_rank_role(member, roles[0])
```
algmyr: Why write this with a comprehension+indexing?
```
if user.rank.title not in ctx.guild.roles:
    raise ...
await self.update_member_rank_role(member, user.rank.title)
```

tle/cogs/handles.py
```diff
@@ -224,109 +228,70 @@ def __init__(self, bot):
...
-                    print(e)
-            msg = 'Update roles completed.'
-        except Exception as e:
-            msg = 'updateroles error!'
-            print(e)
-        await ctx.send(msg)
+    async def updateroles(self, ctx):
+        """Update Codeforces rank roles for everyone."""
+        res = cf_common.user_db.getallhandles()
+        member_handles = [(ctx.guild.get_member(int(user_id)), handle) for user_id, handle in res]
```
algmyr: Nothing to improve this at the moment, but this is where python 3.8 will come in handy with `:=`.

```
member_handles = [(member, handle) for user_id, handle in res
                  if (member := ctx.guild.get_member(int(user_id)))]
```

tle/cogs/handles.py
```diff
@@ -224,109 +228,70 @@ def __init__(self, bot):
...
+        member_handles = [(member, handle) for member, handle in member_handles if member is not None]
```
algmyr: Do we need `is not None`? Feels like if member should work, but that might be less obvious to the reader. At your own discretion.

tle/cogs/handles.py
meooow25: Nice, done!

tle/cogs/handles.py

tle/cogs/handles.py
meooow25: Seems reasonable, done.

tle/cogs/handles.py
meooow25: I cannot do that since `user.rank.title` is `str` but roles are `discord.Role`s.
I do realize the code is not very nice. Please suggest if you have any alternative.

algmyr: Ah, I missed the `.name`. I guess this is fine, I can't come up with a nicer alternative on the spot other than doing a comprehension beforehand to get a set of role names.

tle/cogs/handles.py
meooow25: I feel it's better to be explicit here.

Issue #25, opened 2019-04-25
new command: print problems solved by user (having certain rating and or tag)


krofna:
given handle and rating or tag print a list of problems solved of that rating/tag by that user
another suggestion: print last 10 problems solved by user, including their rating :tourist_think:

krofna:
also print top x highest rated solved by user in/out of contest

PR #81, opened 2019-07-15
Start working on ;stalk (#25)

krofna:
- filter by tag
- filter by rating range

### DIFF
--- a/tle/cogs/codeforces.py
+++ b/tle/cogs/codeforces.py
@@ -32,8 +32,7 @@ class Codeforces(commands.Cog):
             else:
                 tags.append(arg)
 
-        handles = await cf_common.resolve_handles(ctx, self.converter, ('!' + str(ctx.author),))
-        handle = handles[0]
+        handle, = await cf_common.resolve_handles(ctx, self.converter, ('!' + str(ctx.author),))
         user = cf_common.user_db.fetch_cfuser(handle)
         rating = user.rating
         submissions = await cf.user.status(handle=handle)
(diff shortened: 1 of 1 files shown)


tle/cogs/codeforces.py
```diff
@@ -72,6 +72,26 @@ def __init__(self, bot):
             embed.add_field(name='Matched tags', value=tagslist)
         await ctx.send(f'Recommended problem for `{handle}`', embed=embed)
 
+    @commands.command(brief='Print recently solved problems by user')
+    async def stalk(self, ctx, handle: str):
+        handle = handle or '!' + str(ctx.author)
+        handles = await cf_common.resolve_handles(ctx, self.converter, (handle,))
+        handle = handles[0]
```
algmyr: Since you know you'll get an iterable with 1 element, you could do `handle, = await ...`.

tle/cogs/codeforces.py
```diff
@@ -72,6 +72,26 @@ def __init__(self, bot):
...
+        user = cf_common.user_db.fetch_cfuser(handle)
+        submissions = await cf.user.status(handle=handle)
+        submissions = list({sub.problem.name : sub for sub in submissions
+                            if sub.author.participantType == 'PRACTICE'
+                            and sub.verdict == 'OK'}.values())
+        submissions.sort(key=lambda sub: -sub.creationTimeSeconds)
```
algmyr: I know this works, but the pythonic™ way of doing this would probably be
```
submissions.sort(key=lambda sub: sub.creationTimeSeconds, reverse=True)
```

tle/cogs/codeforces.py
```diff
@@ -72,6 +72,26 @@ def __init__(self, bot):
...
+        problems = [sub.problem for sub in submissions]
+
+        msg = ''
+        for i in range(min(5, len(problems))):
```
algmyr: This is equivalent to
```
for prob in problems[:5]:
   ...
```
in this case.
*Edit: see next comment*

tle/cogs/codeforces.py
```diff
@@ -72,6 +72,26 @@ def __init__(self, bot):
...
+            prob = problems[i]
+            rating = prob.rating if prob.rating else '?'
+            msg += f'{prob.name} [{rating}] - <{prob.url}>\n'
+        await ctx.send(msg)
```
algmyr: Actually, maybe even do
```
msg = '\n'.join(
    f'{prob.name} [{prob.rating if prob.rating else "?"}] - <{prob.url}>\n'
    for prob in problems[:5])
```

tle/cogs/codeforces.py
```diff
@@ -72,6 +72,26 @@ def __init__(self, bot):
             embed.add_field(name='Matched tags', value=tagslist)
         await ctx.send(f'Recommended problem for `{handle}`', embed=embed)
 
+    @commands.command(brief='Print recently solved problems by user')
```
algmyr: Maybe mention it's practice problems?

tle/cogs/codeforces.py
krofna: i didnt know that syntax. thanks

algmyr: It's actually just tuple syntax. `a, == (a,)`. In general stuff like
```
a,b = 1,2
```
is just syntactic sugar for
```
(a,b) = (1,2)
```

PR #82, opened 2019-07-15
Add rate limiter

algmyr:
Draft for the rate limiter. It needs to be integrated, and we should consider whether we want the lock as we currently do.

### DIFF
--- a/tle/util/codeforces_api.py
+++ b/tle/util/codeforces_api.py
@@ -165,9 +165,6 @@ class RatingChangesUnavailableError(CodeforcesApiError):
     def __init__(self, comment, contest_id):
         super().__init__(comment, f'Rating changes unavailable for contest with ID `{contest_id}`')
 
-class ClientError(CodeforcesApiError):
-    def __init__(self):
-        super().__init__(None, 'Error connecting to Codeforces API')
 
 # Codeforces API query methods
 
--- a/tle/util/codeforces_api.py
+++ b/tle/util/codeforces_api.py
@@ -168,14 +168,11 @@ class RatingChangesUnavailableError(CodeforcesApiError):
 # Codeforces API query methods
 
 _session = None
-_api_lock = None
 
 
 async def initialize():
     global _session
-    global _api_lock
     _session = aiohttp.ClientSession()
-    _api_lock = asyncio.Lock()
 
 
 def _bool_to_str(value):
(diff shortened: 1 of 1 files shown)


tle/util/codeforces_api.py
```diff
@@ -183,30 +181,59 @@ def _bool_to_str(value):
...
+    async def wrapped(*args, **kwargs):
+        tries = 3
+        for i in range(tries):
+            now = time.time()
+
+            # Next valid slot is 1s after the `per_second`th last request
+            next_valid = max(now, 1 + last[0])
+            last.append(next_valid)
+            last.popleft()
+
```
cheran-senthil: A [queue](https://docs.python.org/3.7/library/queue.html) be a more accurate structure here.

algmyr: queue is for when synchronization is needed. IIRC for deque the relevant operations are atomic which makes the usage here fine, no need for synchronization.

tle/util/codeforces_api.py
```diff
@@ -183,30 +181,59 @@ def _bool_to_str(value):
     raise TypeError(f'Expected bool, got {value} of type {type(value)}')
 
 
+def cf_ratelimit(f):
+    per_second = 5
+    last = deque([0]*per_second)
+    async def wrapped(*args, **kwargs):
```
meooow25: It's a good practice to use functools.wraps.

tle/util/codeforces_api.py
```diff
@@ -183,30 +181,59 @@ def _bool_to_str(value):
...
+            if delay > 0:
+                await asyncio.sleep(delay)
+
+            try:
+                return await f(*args, **kwargs)
+            except (ClientError, CallLimitExceededError) as e:
+                if i < tries - 1:
+                    logger.info(f'Try {i+1}/{tries} at query failed. Retrying...')
+                else:
+                    logger.info(f'Try {i+1}/{tries} at query failed. Aborting.')
```
meooow25: We should log the error repr too. Also both messages are similar, we can print the common part once and "Retrying/Aborting" separately.

tle/util/codeforces_api.py
```diff
@@ -183,30 +181,59 @@ def _bool_to_str(value):
...
+                    raise e
```
meooow25: Shouldn't there be a delay between tries? Especially when it's CallLimitExceeded, retrying immediately makes things worse.

algmyr: So, add some artificial delay if it's not the first instance of the loop?

tle/util/codeforces_api.py
algmyr: I had no idea this was a thing, good to know.

algmyr:
The discussions regarding this in a nutshell:
This is probably the simplest rate limiter that should do the job if CF respects their own call limits.
It will also behave nicely in an outage (failing quickly).
Adding stuff to it only means messing the idea of it up.
We can replace it with a better system later if needed, but this will do for now at least.

Issue #47, opened 2019-05-16
plot performance like ;plot rating


krofna:

krofna:
precalculate performances for all users in all contests and store in cache.
see #43 #196

Aaerialys:
this shouldn't be too hard right? just approximate it as current rating + delta*4

krofna:
thats not performance. thats rating + delta*4

Aaerialys:
yes but it is close to the performance for most circumstances and easy to add to the current ;ranklist command because you already know the delta and rating of each person

for example rating+delta*4 is used to calculate performance here: https://greasyfork.org/en/scripts/402180-codeforces-performance

krofna:
I mean actual performance, as calculated by carrot addon.

PR #83, opened 2019-07-15
Start working on performance plot (#47)

krofna:
This is a draft.
- ugly code
- incredibly slow (takes couple of minutes)
- tends to crash (API limit exceeded)
- code duplication


krofna:
@meooow25 this is wrong. i need to plot which rating would get +0 and this obviously isn't it

meooow25:
We should develop a new cache for this since a single command will query potentially > 50 contests.

PR #84, opened 2019-07-15
On gotgud, display the time it took (#45)

krofna:


tle/cogs/codeforces.py
```diff
@@ -152,12 +152,25 @@ def check(problem):
             await ctx.send('You haven\'t completed your challenge.')
             return
 
+        def pretty_time_delta(seconds):
```
algmyr: Pedantry: singular/plural is not handled properly.
To not be profusely annoying, here's working code
```
def pretty_time_delta(seconds):
    seconds = int(seconds)
    days, seconds = divmod(seconds, 86400)
    hours, seconds = divmod(seconds, 3600)
    minutes, seconds = divmod(seconds, 60)

    timespec = [
        (days, 'day', 'days'),
        (hours, 'hour', 'hours'),
        (minutes, 'minute', 'minutes'),
        #(seconds, 'second', 'seconds'),
    ]

    return ' '.join(f'{count} {singular if count == 1 else plural}'
                    for count, singular, plural in timespec)
```

PR #85, opened 2019-07-15
Only discard problem from gimme if its actually AC

krofna:
maybe do same with gitgud?


algmyr:
Yeah, this should be done with gitgud as well. Make another PR for that.

PR #95, opened 2019-07-29
added cses_individual

pantsudestroyer:


tle/cogs/cses.py
```diff
@@ -76,6 +76,41 @@ def leaderboard(self, placings, num):
 
         return str(t)
 
+    def leaderboard_individual(self, placings, *handles: str):
+        leaderboard = sorted(
+            ((k, score(v)) for k, v in placings.items() if k != 'N/A' and k in handles),
+            key=lambda x: x[1],
+            reverse=True)
+        
+        _placings = defaultdict(list)
```
algmyr: Why can't you just use `placings` as is at line 109? As far as I see `_placings` just contains the same info but for fewer users.

tle/cogs/cses.py
```diff
@@ -76,6 +76,41 @@ def leaderboard(self, placings, num):
...
+        included = [handle for handle, score in leaderboard]
+        for handle in handles:
+            if handle not in included:
+                leaderboard.append((handle, 0))
+            else:
+                _placings[handle] = placings[handle]
```
algmyr: If `_placings` is removed (see above comment) then this could be shortened a lot. 87-91 becomes just
```
leaderboard += [(handle, 0) for handle in handles if handle not in included]
```
(haven't tested the code, so please do a quick run to check)
**NB.** python trivia += on a list is *exactly* the same as .extend

tle/cogs/cses.py
```diff
@@ -76,6 +76,41 @@ def leaderboard(self, placings, num):
...
+        
+        if not leaderboard:
```
algmyr: 93-112 is shared code, right? If so we could split it off into an helper function and use it in both places, like `format_leaderboard(top, placings)`

tle/cogs/cses.py
```diff
@@ -84,11 +119,26 @@ def fastest(self, num=10):
...
+    def shortest_individual(self, *handles: str):
+        return self.leaderboard_individual(self.short_placings, *handles)
+
     @commands.command(brief='Shows compiled CSES leaderboard')
     async def cses(self, ctx):
         """Shows compiled CSES leaderboard."""
         await ctx.send('```\n' 'Fastest\n' + self.fastest + '\n\n' + 'Shortest\n' + self.shortest + '\n' + '```')
 
+    @commands.command(brief='Shows compiled CSES leaderboard for individuals')
+    async def cses_individual(self, ctx, *handles: str):
```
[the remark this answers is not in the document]
algmyr: Also, at the merge you should add/update usage notes and similar, see https://github.com/cheran-senthil/TLE/blob/dda3dda106c9d66172200f00f3ee7ecaa7e12ab7/tle/cogs/graphs.py#L393 for an example.

The things to cover:
* brief will contain a very brief explanation, say "Shows CSES leaderboard"
* usage would show the user than arguments can be supplied (optionally), similar to in the case I linked
* the docstring after the function declaration can be expanded to explain a bit more in depth what happens with/without arguments, again see the linked code for an example

PR #96, opened 2019-08-03
Improve ;vc

krofna:


tle/cogs/codeforces.py
```diff
@@ -294,18 +294,30 @@ def pretty_time_delta(seconds):
             await ctx.send(f'Failed to force challenge skip.')
 
     @commands.command(brief='Recommend a contest')
-    async def vc(self, ctx, *handles: str):
+    async def vc(self, ctx, *args: str):
         """Recommends a contest based on Codeforces rating of the handle provided."""
-        handles = handles or ('!' + str(ctx.author),)
+        args = list(args)
```
algmyr: this is not needed if you go the route mentioned below without the remove

PR #101, opened 2019-08-05
Identify

krofna:


tle/cogs/handles.py
```diff
@@ -146,10 +150,11 @@ def __init__(self, bot):
     @commands.has_role('Admin')
     async def set(self, ctx, member: discord.Member, handle: str):
         """Set Codeforces handle of a user."""
+        # CF API returns correct handle ignoring case, update to it
         users = await cf.user.info(handles=[handle])
-        user = users[0]
+        await self._set(ctx, member, users[0])
```
algmyr: I think this a typical case where you would write a decorator, but this is fine.

tle/cogs/handles.py
```diff
@@ -163,6 +168,35 @@ def __init__(self, bot):
...
+            return
+
+        handle, prob, url = self.identify_map[invoker]
+        subs = await cf.user.status(handle=handle, count=5)
+        if any(sub.problem.name == prob and sub.verdict == 'COMPILATION_ERROR' for sub in subs):
+            self.identify_map.pop(invoker)
+            users = await cf.user.info(handles=[handle])
+            await self._set(ctx, ctx.author, users[0])
+        else:
+            await ctx.send(f'Sorry `{invoker}`, can you try again?')
```
algmyr: General note: We probably should write a function that escapes usernames so we actually print the correct thing. Something that prepends <code>*_`</code> with a backslash would do the job.

tle/cogs/handles.py
```diff
@@ -163,6 +168,35 @@ def __init__(self, bot):
...
+    @handle.command(brief='Identify yourself')
+    async def identify(self, ctx, handle: str):
+        users = await cf.user.info(handles=[handle])
+        if not users:
+            await ctx.send('`{handle}` not found on codeforces')
+            return
+
+        invoker = str(ctx.author)
+        problem = random.choice(cf_common.cache2.problem_cache.problems)
+        handle, name, url = (users[0].handle, problem.name, problem.url)
```
algmyr: This feels a bit superfluous considering most of these are just used once. Just write them in-place, the names are obvious enough and not too long.

tle/cogs/handles.py
```diff
@@ -163,6 +168,35 @@ def __init__(self, bot):
...
+    async def report(self, ctx):
+        invoker = str(ctx.author)
+        if not invoker in self.identify_map:
+            await ctx.send(f'`{invoker}`, you have nothing to report')
+            return
+
+        handle, prob, url = self.identify_map[invoker]
+        subs = await cf.user.status(handle=handle, count=5)
+        if any(sub.problem.name == prob and sub.verdict == 'COMPILATION_ERROR' for sub in subs):
+            self.identify_map.pop(invoker)
```
algmyr: The general wisdom for removing things from dicts in python is to use `value = D.pop(key)` if you want to capture the value, otherwise just do `del D[key]`.

tle/cogs/handles.py
```diff
@@ -163,6 +168,35 @@ def __init__(self, bot):
...
+    async def identify(self, ctx, handle: str):
+        users = await cf.user.info(handles=[handle])
+        if not users:
+            await ctx.send('`{handle}` not found on codeforces')
+            return
+
+        invoker = str(ctx.author)
+        problem = random.choice(cf_common.cache2.problem_cache.problems)
+        handle, name, url = (users[0].handle, problem.name, problem.url)
+        self.identify_map[invoker] = (handle, name, url)
```
algmyr: Do you need to store the url? It's not used for anything right now afaik.

tle/cogs/handles.py
```diff
@@ -163,6 +168,35 @@ def __init__(self, bot):
             raise HandleCogError(f'Role for rank `{user.rank.title}` not present in the server')
         await self.update_member_rank_role(member, roles[0])
 
+    @handle.command(brief='Identify yourself')
+    async def identify(self, ctx, handle: str):
```
algmyr: Docstring and usage?

Issue #74, opened 2019-07-08
allow filter by tags like +dp in ;solved


krofna:

algmyr:
Closed by #105

PR #105, opened 2019-08-08
Added tags functionality to ;plot solved

robert-30:


robert-30:
implements #74

tle/cogs/graphs.py
```diff
@@ -72,18 +72,22 @@ def _plot_rating(resp, mark='o', labels: List[str] = None):
     _plot_rating_bg()
 
 
-def _filter_solved_submissions(submissions, contests):
+def _filter_solved_submissions(submissions, contests, *args):
```
algmyr: Shouldn't the last argument just be something like `tags = []`, not `*args`?

tle/cogs/graphs.py
```diff
@@ -72,18 +72,22 @@ def _plot_rating(resp, mark='o', labels: List[str] = None):
...
+    to filter problems. If a problem is solved multiple times the first accepted
     submission is kept. The unique id for a problem is (problem name, contest start time).
     """
     submissions.sort(key=lambda sub: sub.creationTimeSeconds)
     contest_id_map = {contest.id: contest for contest in contests}
     problems = set()
     solved_subs = []
+    tags = []
+    if (len(args) > 0):
+        tags = args[0]
```
algmyr: This can be entirely removed if the signature is fixed as described above.

tle/cogs/graphs.py
```diff
@@ -72,18 +72,22 @@ def _plot_rating(resp, mark='o', labels: List[str] = None):
...
     for submission in submissions:
         contest = contest_id_map.get(submission.problem.contestId)
-        if submission.verdict == 'OK' and submission.problem.rating and contest:
+        if submission.verdict == 'OK' and submission.problem.rating and contest and (tags == [] or submission.problem.tag_matches(tags)):
```
algmyr: Do `not tags`. It will work also in the case where a tuple is given as tags.

robert-30:
Updated

PR #115, opened 2019-08-19
Add support for user_guard error

meooow25:
user_guard raises the optionally provided exception when the group is active.

This is not a great system but better than silently ignoring.


krofna:
@meooow25 why did this PR die? It looks good. I'll merge this today if no complaints arrive.

meooow25:
I totally forgot about this 👀
It's quite minor anyway.

PR #119, opened 2019-08-24
Adding logging to file

### LINKED ISSUE #54: Set up logging to a file
Currently the bot logs to console. Configure it to log to a file as well as console.

[howto 1](https://docs.python.org/3.7/howto/logging.html)
[howto 2](https://docs.python.org/3.7/howto/logging-cookbook.html)

kabeer27:
Added logging to a file while maintaining logging to console

Resolves #54.

### DIFF
--- a/tle/__main__.py
+++ b/tle/__main__.py
@@ -19,9 +19,7 @@ def setup():
     logging.basicConfig(format='{asctime}:{levelname}:{name}:{message}', style='{',
                         datefmt='%d-%m-%Y %H:%M:%S', level=logging.INFO,
                         handlers=[logging.StreamHandler(),
-                                  TimedRotatingFileHandler('TleBot.log',when='D',utc=True)
-                                  ]
-                        )
+                                  TimedRotatingFileHandler('TleBot.log', when='D', utc=True)])
 
     # matplotlib and seaborn
     plt.rcParams['figure.figsize'] = 7.0, 3.5
--- a/tle/__main__.py
+++ b/tle/__main__.py
@@ -14,9 +14,11 @@ from tle.util import discord_common
 
 
 def setup():
-    # logging
-    logging.basicConfig(format='{asctime}:{levelname}:{name}:{message}', style='{',
+    # logging to file
+    logging.basicConfig(filename="log.txt",filemode="a",format='{asctime}:{levelname}:{name}:{message}', style='{',
                         datefmt='%d-%m-%Y %H:%M:%S', level=logging.INFO)
+    # logging to console
+    logging.getLogger().addHandler(logging.StreamHandler())
 
     # matplotlib and seaborn
     plt.rcParams['figure.figsize'] = 7.0, 3.5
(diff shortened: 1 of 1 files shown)


tle/__main__.py
```diff
@@ -14,9 +14,11 @@
 
 
 def setup():
-    # logging
-    logging.basicConfig(format='{asctime}:{levelname}:{name}:{message}', style='{',
+    # logging to file
+    logging.basicConfig(filename="log.txt",filemode="a",format='{asctime}:{levelname}:{name}:{message}', style='{',
                         datefmt='%d-%m-%Y %H:%M:%S', level=logging.INFO)
+    # logging to console
+    logging.getLogger().addHandler(logging.StreamHandler())
```
meooow25: This handler will use the default formatting instead of the one for the file.
It is also not very clear what is happening here, which is basicConfig creating a FileHandler and a StreamHandler being added after that.

You can instead explicitly pass the two handlers to the `handlers` parameter of basicConfig. Also I am now aware of something called TimedRotatingFileHandler and I suggest we use that for logging to disk.

algmyr:
Closes #54

tle/__main__.py
```diff
@@ -1,5 +1,6 @@
 import argparse
 import logging
+from logging.handlers import TimedRotatingFileHandler
```
meooow25: Minor detail: move this down to after `import os`.

tle/__main__.py
```diff
@@ -14,9 +15,13 @@
...
 def setup():
-    # logging
+    # logging to console and File on Daily interval
     logging.basicConfig(format='{asctime}:{levelname}:{name}:{message}', style='{',
-                        datefmt='%d-%m-%Y %H:%M:%S', level=logging.INFO)
+                        datefmt='%d-%m-%Y %H:%M:%S', level=logging.INFO,
+                        handlers=[logging.StreamHandler(),
+                                  TimedRotatingFileHandler("TleBot.log",when="D",interval=1,backupCount=0,utc=True)
+                                  ]
+                        )
```
meooow25:
```py
logging.basicConfig(format='{asctime}:{levelname}:{name}:{message}', style='{',
                    datefmt='%d-%m-%Y %H:%M:%S', level=logging.INFO,
                    handlers=[logging.StreamHandler(),
                              TimedRotatingFileHandler('tle.log', when='D', utc=True)])
```

- It's unnecessary to specify default values for parameters (`interval`, `backupCount`).
- Use ' strings for consistency with this project.

tle/__main__.py
```diff
@@ -14,9 +15,13 @@
 
 
 def setup():
-    # logging
+    # logging to console and File on Daily interval
```
meooow25: Unnecessary caps, File -> file, Daily -> daily.

tle/__main__.py
meooow25: Style issues: space between parameters and closing braces on the the same line (see snippet above).

PR #125, opened 2019-08-26
Add spacing between the numbers and the usernames

kaguwuya:


algmyr:
Sure this solves the problem, but it would be nice if it was actually done dynamically. I.e. check the lengths of all numbers that will be shown and pad accordingly or something equivalent.

Issue #133, opened 2019-08-31
;plot howgud without arguments is broken


krofna:

krofna:
should print error instead

algmyr:
Closed by #134

PR #134, opened 2019-09-01
Fix for Issue #133

aneesh2312:
#133 
If no arguments specified, use the message author's id


tle/cogs/graphs.py
```diff
@@ -602,6 +602,8 @@ def vert_line(x):
 
     @plot.command(brief='Plot histogram of gudgiting')
     async def howgud(self, ctx, *members: discord.Member):
+        if len(members) == 0:
+            members += (ctx.author,)
```
algmyr: You can do
```
members = members or (ctx.author,)
```
if you want a one-liner. Either change or don't, then just merge this. Such a minor code change that waiting for reviews is a bit silly.

PR #153, opened 2019-09-24
Correctly manage roles in certain situations

### LINKED ISSUE #144: Setting handle to unrated user doesn't remove previously set rank role

meooow25:
...where the certain situations are
- Handle change from rated to unrated user
- Handle removal
- Role update caused by `;updateroles`

Fixes #144.

### DIFF
--- a/tle/cogs/handles.py
+++ b/tle/cogs/handles.py
@@ -165,8 +165,6 @@ class Handles(commands.Cog):
         handle = user.handle
         cf_common.user_db.cache_cfuser(user)
         cf_common.user_db.sethandle(member.id, handle)
-        embed = _make_profile_embed(member, user, mode='set')
-        await ctx.send(embed=embed)
 
         if user.rank == cf.UNRATED_RANK:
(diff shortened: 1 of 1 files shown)


tle/cogs/handles.py
```diff
@@ -207,6 +217,8 @@ def __init__(self, bot):
             raise HandleCogError(f'Handle for {member.mention} not found in database')
         embed = discord_common.embed_success(f'Removed handle for {member.mention}')
         await ctx.send(embed=embed)
+        await self.update_member_rank_role(member, role_to_assign=None,
+                                           reason='Handle removed for user')
```
algmyr: I think it would be more reasonable to have

    embed = ...
    await ctx.send...

at the end when everything is actually done.

meooow25: `set` is similar actually. I consider that the first step is the actual set/removal since it modifies the db, and the role update is secondary. Agree it's not very nice though. Do you still think we should change it?

PR #158, opened 2019-09-27
pretty handle support more fonts

### LINKED ISSUE #150: ;handle pretty doesn't support japanese

goswami-rahul:
- Now supports CJK fonts
- No more monospace fonts
- Looks nicer
- Resolves #150 

image
image


tle/cogs/handles.py
```diff
@@ -56,32 +56,49 @@ def get_prettyhandles_image(rankings):
...
-        s = f"{pos:<4}{name:<18}{handle:<18}{rating:>6}"
+    START_X, START_Y = 20, 20
+    Y_INC = 32
+    WIDTH_RANK = 64
+    WIDTH_NAME = 320
+
+    x, y = START_X, START_Y
+
+    def draw_row(pos, username, handle, rating, color):
+        nonlocal x, y
```
meooow25: Persisting nonlocal x and y isn't really necessary here? `x` is always `START_X` and `y` can be passed to the function.

tle/cogs/handles.py
```diff
@@ -56,32 +56,49 @@ def get_prettyhandles_image(rankings):
...
+
+    # draw header
+    draw_row('#', 'Username', 'Handle', 'Rating', BLACK)
+    y += int(Y_INC * 0.5)
+
+    # trim name to fit in the column width
+    def _trim(name):
+        width = WIDTH_NAME - 10
+        while font.getsize(name)[0] > width:
+            name = name[:-3] + '..'
```
meooow25: Use the ellipsis character `…`.

tle/cogs/handles.py
```diff
@@ -56,32 +56,49 @@ def get_prettyhandles_image(rankings):
...
-            draw.text((z, y), s[22], fill=BLACK, font=font)
-            z += font.getsize((s[22]))[0]
-            draw.text((z, y), s[23:], fill=color, font=font)
-        else:
-            draw.text((x, y), s, fill=color, font=font)
-        y += y_inc
+        draw_row(str(pos), name, handle, str(rating), color)
+        if rating != 'N/A' and rating >= 3000:  # nutella
+            nutella_x, nutella_y = x + WIDTH_RANK, y - Y_INC
+            draw.text((nutella_x, nutella_y), name[0], fill=BLACK, font=font)
```
meooow25: Nice way to handle nutella.

tle/cogs/handles.py
goswami-rahul: We switched to ellipsis when we used monospace, to save some width. But now ellipsis … takes more width than two full stops .. with this font. And the ellipsis is printed as middle height dots in this font. Maybe we can use 3 full stops instead of 2.
image

meooow25: Looks like it's an issue with the font that horizontal ellipsis U+2026 "…" is displayed the same as midline horizontal ellipsis U+22EF "⋯". For Noto Sans (non-CJK) this does not happen, but it cannot display CJK characters :(

Let's just use three dots then, I guess.

algmyr: With some extra legwork you can mix the fonts. Shouldn't be too horrible considering we only intend to treat the … special. In short, draw the text without the ellipsis at first then use `PIL.ImageDraw.Draw.textsize` to compute the dimensions and then draw the ellipsis offset by the computed horizontal dimension.

tle/cogs/handles.py
```diff
@@ -16,6 +18,9 @@
 _HANDLES_PER_PAGE = 15
 _NAME_MAX_LEN = 20
 _PAGINATE_WAIT_TIME = 5 * 60  # 5 minutes
+_FONT_PATH = "tle/assets/fonts/NotoSansCJK-Bold.ttc"
```
meooow25: Let's move it to `tle/constants.py` with a more descriptive name.

tle/cogs/handles.py
```diff
@@ -55,33 +60,48 @@ def get_prettyhandles_image(rankings):
...
-    header = f"{'#':<4}{'Username':<18}{'Handle':<18}{'Rating':>7}"
-    draw.text((x, y), header, fill=BLACK, font=font)
-    y += int(y_inc * 1.5)
-    for pos, name, handle, rating in rankings:
-        if len(name) > 17:
-            name = name[:16] + "…"
-        if len(handle) > 17:
-            handle = handle[:16] + "…"
-        s = f"{pos:<4}{name:<18}{handle:<18}{rating:>6}"
+    font = ImageFont.truetype(_FONT_PATH, size=26)
```
meooow25: This can be moved inside the cog and loaded only once on startup.

tle/cogs/handles.py
```diff
@@ -252,6 +272,10 @@ def __init__(self, bot):
 
     @handle.command(brief="Show colour handles")
     async def pretty(self, ctx: discord.ext.commands.Context, page_no: int = None):
+        if not os.path.isfile(_FONT_PATH):
```
meooow25: Let's check only once on startup.
```py
class Handles(commands.Cog):
    def __init__(self, bot):
        self.bot = bot
        self.font = None
        self.logger = logging.getLogger(self.__class__.__name__)

    @commands.Cog.listener()
    async def on_ready(self):
        try:
            self.font = ImageFont.truetype(constants.NOTO_SANS_CJK_FONT_PATH, size=26)
        except OSError:
            url = 'https://noto-website-2.storage.googleapis.com/pkgs/NotoSansCJK-Bold.ttc.zip'
            self.logger.warning(f'Font file {constants.NOTO_SANS_CJK_FONT_PATH} not found. Pretty '
                                f'handles is disabled. Download from {url}.')
            self.handle.remove_command(self.pretty.name)
...
```

tle/cogs/handles.py
algmyr: I think you might be slightly damaged by work.

goswami-rahul: :sadness:

PR #159, opened 2019-09-27
Refactor caching problemsets

### LINKED ISSUE #139: problemset cache is broken
we save it before mike puts up ratings and don't update it

meooow25:
Refactored how problemsets are cached.

- Add a dedicated `ProblemsetCache` class to cache problem sets.
- This class keeps monitoring finished contests until their problems get assigned ratings, for upto 14 days.

Fixes #139.

### DIFF
--- a/tle/cogs/cache_control.py
+++ b/tle/cogs/cache_control.py
@@ -72,15 +72,14 @@ class CacheControl(commands.Cog):
         """Mode 'all' clears all existing cached problems. Mode 'contest_id'
         clears existing problems with the given contest id.
         """
-        if contest_id != 'all':
-            try:
-                contest_id = int(contest_id)
-            except ValueError:
-                return
         if contest_id == 'all':
             await ctx.send('This will take a while')
             count = await cf_common.cache2.problemset_cache.update_for_all()
         else:
+            try:
+                contest_id = int(contest_id)
+            except ValueError:
+                return
             count = await cf_common.cache2.problemset_cache.update_for_contest(contest_id)
         await ctx.send(f'Done, fetched {count} problems')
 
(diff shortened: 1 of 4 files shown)


tle/cogs/cache_control.py
```diff
@@ -63,8 +65,28 @@ def __init__(self, bot):
...
+        if contest_id != 'all':
+            try:
+                contest_id = int(contest_id)
+            except ValueError:
+                return
+        if contest_id == 'all':
+            await ctx.send('This will take a while')
+            count = await cf_common.cache2.problemset_cache.update_for_all()
+        else:
+            count = await cf_common.cache2.problemset_cache.update_for_contest(contest_id)
```
algmyr: Rewrite this as a single if-else. So move L84 up and make it an else.

tle/util/cache_system2.py
```diff
@@ -243,6 +235,93 @@ def keep(problem):
...
+            new_problems, updated_problems = await self._fetch_problemsets(contests)
+            self._save_problems(new_problems + updated_problems)
+            self.logger.info(f'{len(new_problems)} new problems saved and {len(updated_problems)} '
+                             'saved problems updated.')
+
+    async def _fetch_problemsets(self, contests, *, force_fetch=False):
+        # We assume it is possible for problems in the same contest to get assigned rating at
+        # different times.
+        new_contest_ids = []
+        saved_contests = {}
```
algmyr: This name could be more descriptive.

tle/cogs/cache_control.py
meooow25: Makes sense, done.

tle/util/cache_system2.py
meooow25: Updated to a list of pairs with a better name.

Issue #156, opened 2019-09-26
Add ping stats to ;ping


meooow25:
Show time taken.

PR #163, opened 2019-10-01
Ping command now shows ping, issue #156

albertopoljak:


meooow25:
Not exactly sure this is what we want. This would be the time taken to send an HTTP request to Discord API. Will investigate and update here tomorrow.

meooow25:
This change is alright, if you could also add the websocket/gateway latency the PR is good to go. This can be obtained as `bot.latency`. The message should look like:

**REST API latency**: 345ms
**Gateway API latency**: 678ms

tle/cogs/meta.py
```diff
@@ -69,7 +70,12 @@ def __init__(self, bot):
     @meta.command(brief='Is TLE up?')
     async def ping(self, ctx):
         """Replies to a ping."""
-        await ctx.send("pong")
+        start = time.perf_counter()
+        message = await ctx.send(':ping_pong: Pong!')
+        end = time.perf_counter()
+        duration = (end - start) * 1000
+        await message.edit(content=f'REST API latency: {int(duration)}ms \n'
```
krofna: The extra whitespace before the newline is slightly frustrating.

albertopoljak: Ayy sorry. 
Little design issues like this are hardest.
At first I had whitespace but `{int(duration)}ms\n` just seemed like too much special characters without a break so I added whitespace for easier reading I guess.
Reverted back now.

tle/cogs/meta.py
```diff
@@ -69,7 +70,12 @@ def __init__(self, bot):
     @meta.command(brief='Is TLE up?')
     async def ping(self, ctx):
         """Replies to a ping."""
-        await ctx.send("pong")
+        start = time.perf_counter()
+        message = await ctx.send(':ping_pong: Pong!')
```
meooow25: `:ping_pong:` seems unnecessary. Also it is removed in the edit anyway.

krofna:
I'm merging this. Thank you for your contribution.

PR #173, opened 2019-10-04
Add timestamp to ;stalk

algmyr:
Untested, am on very limited internet connection.

### DIFF
--- a/tle/util/codeforces_common.py
+++ b/tle/util/codeforces_common.py
@@ -1,7 +1,9 @@
 import functools
 import json
 import logging
+import math
 import os
+import time
 from collections import defaultdict
 
 from discord.ext import commands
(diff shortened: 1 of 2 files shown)


tle/util/codeforces_common.py
```diff
@@ -171,6 +173,14 @@ def pretty_time_format(seconds):
     return ' '.join(f'{count} {singular if count == 1 else plural}'
             for count, singular, plural in timeprint)
 
+def days_ago(t):
```
meooow25: This is rather specific and might just live inside stalk command? Upto you.

algmyr: Like discussed elsewhere, it was moved on request

algmyr:
I'll merge this for now, if we feel the need to improve how we display the information we can do that later on.

PR #177, opened 2019-10-06
At least alleviate the issues with wide chars

### LINKED ISSUE #136: Make table code handle wide symbols
Some symbols take up two monospace characters, handle that.

Try using this: https://pypi.org/project/wcwidth/. In short, I think the two calls to `len` in https://github.com/cheran-senthil/TLE/blob/master/tle/util/table.py should just be replaced.

algmyr:
Too bad the code environment in discord is not proper monospace for
whatever reason, so this will never be absolutely correct.

Closes #136 to the best of my ability given what we have to work with.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,6 @@ python = "^3.7"
 seaborn = "^0.9.0"
 lxml = "^4.3"
 pillow = "^5.4"
-wcwidth = "^0.1.7"
 
 [tool.poetry.dev-dependencies]
 pytest = "^3.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,6 +10,7 @@ python = "^3.7"
 seaborn = "^0.9.0"
 lxml = "^4.3"
 pillow = "^5.4"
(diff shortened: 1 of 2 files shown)


algmyr:
Err, this needs more work. I'll re-open when I get it working.

algmyr:
Ok, it works now. Vast improvement compared to what is was, but still broken because discord.
image

meooow25:
Looks decent. Can we avoid the dependency on wcwidth using [unicodedata.east_asian_width](https://docs.python.org/3.7/library/unicodedata.html#unicodedata.east_asian_width)?

algmyr:
Possibly, but `east_asian_width` would require some wrapping function to make it handle full strings, and also to convert the descriptions it returns (Narrow, Full width, ...) to the actual widths. I could easily do it, but is it worthwhile compared to just accepting the dependency?

algmyr:
This is now much improved by not assuming full width chars are 2 wide, currently we're using 1.66667 which seems to work good for most users.

PR #178, opened 2019-10-07
Adds some silly presence messages

algmyr:
First message just shows TLE is ready to receive commands. After that it's just orzing random users.

This could be improved later on to do other things, but it felt like wasted potential to have nothing there.

### DIFF
--- a/tle/__main__.py
+++ b/tle/__main__.py
@@ -69,6 +69,8 @@ def main():
     @bot.event
     async def on_ready():
         await cf_common.initialize(args.nodb)
+        await discord_common.presence(bot)
+        
 
     bot.add_listener(discord_common.bot_error_handler, name='on_command_error')
 
--- a/tle/__main__.py
+++ b/tle/__main__.py
@@ -1,3 +1,4 @@
+import asyncio
 import argparse
 import logging
 import os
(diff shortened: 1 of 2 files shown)


tle/util/discord_common.py
```diff
@@ -54,3 +55,13 @@ def set_author_footer(embed, user):
     else:
         exc_info = type(exception), exception, exception.__traceback__
         logger.exception('Ignoring exception in command {}:'.format(ctx.command), exc_info=exc_info)
+
+
+async def presence(bot):
+    await bot.change_presence(activity=discord.Game(
+        name="I'm ready for inputs!"))
+    while True:
+        await asyncio.sleep(60)
```
meooow25: A minute seems too low. Maybe 10?

tle/__main__.py
```diff
@@ -65,6 +65,8 @@ def no_dm_check(ctx):
     @bot.event
     async def on_ready():
         await cf_common.initialize(args.nodb)
+        await discord_common.presence(bot)
```
meooow25: The listener gets stuck in an infinite loop. Not sure if it affects discord.py but it's not very nice either way. Use `asyncio.create_task` instead.

tle/util/discord_common.py
```diff
@@ -54,3 +55,13 @@ def set_author_footer(embed, user):
...
+        target = random.choice(bot.users)
```
krofna: [user for user in bot.users if purgatory not in user.roles]

PR #181, opened 2019-10-07
Support for CJK in howgud. Closes #165

### LINKED ISSUE #165: Add CJK font support to matplotlib plots
..such as in this label

image

Also see #150.

krofna:
**This is a draft.**
- why were we prepending that zero width space? doesn't work nice with the font we use for pretty handles. 
- bold doesn't look nice i think

image


meooow25:
> why were we prepending that zero width space?

Explanation [here](https://github.com/cheran-senthil/TLE/blob/55746879d20d62a23af40d9f11dfc894c153c1fd/tle/cogs/graphs.py#L375). What happens when you use it with Noto? There should be no effect.

> bold doesn't look nice i think

Yeah, let's use the regular version instead.

tle/cogs/graphs.py
```diff
@@ -643,15 +644,15 @@ def vert_line(x):
...
+        labels = [f'\0{member.display_name}: {len(delta)}'
                   for member, delta in zip(members, deltas)]
 
         plt.clf()
         plt.margins(x=0)
         plt.hist(deltas, bins=hist_bins, label=labels, rwidth=1)
         plt.xlabel('Problem delta')
         plt.ylabel('Number solved')
-        plt.legend()
+        plt.legend(prop=fm.FontProperties(fname=constants.NOTO_SANS_CJK_REGULAR_FONT_PATH))
```
meooow25: We can prepare the font properties just once in `__init__`.

PR #198, opened 2019-10-19
Organize directories

meooow25:
Currently databases are in `./files`, fonts are in `./tle/assets/fonts` and logs are generated in `.` itself.

The new tree would look like:
```
.
├── data
│   ├── assets
│   │   └── fonts
│   │       ├── NotoSansCJK-Bold.ttc
│   │       └── NotoSansCJK-Regular.ttc
│   ├── db
│   │   ├── cache.db
│   │   └── user.db
│   ├── misc
│   │   └── contest_writers.json
│   └── temp
│       └── tempplot_timestamp.png
├── logs
│   ├── tle.log
│   ├── tle.log.timestamp1
│   ├── tle.log.timestamp2
│   └── tle.log.timestamp3
├── tle...
└── extra...
```


.gitignore
```diff
@@ -113,5 +113,8 @@ pip-wheel-metadata/
 # run
 run.sh
 
-# font files
-*.ttc
+# data
+/data
+
+# logs
+/logs
```
algmyr: Add newline at end of file for consistency?

tle/__main__.py
```diff
@@ -17,11 +17,18 @@
 
 
 def setup():
+    # Make required directories.
+    for attribute in dir(constants):
+        if attribute.endswith('DIR'):
+            path = getattr(constants, attribute)
```
algmyr: Not a huge fan of this. Could we make the constants module more structured to avoid this? I propose having a map and then exposing it as a namedtuple:
```
_DIRS = {'dir1': 'path1', 'dir2': 'path2'}
dirs = namedtuple('Directories', _DIRS.keys())(*_DIRS.values())
```
This simplifies the above code to a single simple loop, and element access is still simple `constants.dirs.FONTS` or whatever naming scheme you choose.

algmyr: Could also do the same thing to other natural groups of constants.

meooow25: I'm assuming the `endswith('DIR')` hack is what you refer to? If so I agree it's not too great 👀 

Btw this code could be improved somewhat to
```py
for attrib_name, attrib_value in vars(constants).items():
    if attrib_name.endswith('DIR'):
        os.makedirs(attrib_value , exist_ok=True)
```

To be honest I am not quite in favour of adding a `.dirs` to access the values :thonk:, and I also feel it's a little too complex for the purpose.
An alternate would be to define `DIRS = (DIR_NAME_1, DIR_NAME_2, ...)` but it does bring duplication with it. What do you think?

PR #200, opened 2019-10-22
Fullsolve

kabeer27:
The try block from line 397 to line 404 is added to take care of weird contests like Technocup which do not share their problemsets till a better workaround is found


tle/cogs/codeforces.py
```diff
@@ -377,7 +378,52 @@ def strfilt(s):
...
+
+        full_solve = sorted(packed_contest_subs_problemset, key=lambda contest: (contest[2] - contest[1]))
+        full_solve = [(contest_id, solved, total) for contest_id, solved, total in full_solve if (total - solved) > 0]
+
+        if not full_solve:
+            await ctx.send('User has no contests to fullsolve')
+        else:
+            def make_line(entry):
+                contest, solved, total = entry
+                line = f'[{contest.name}]({contest.url})\N{EN SPACE}[{solved}/{total}]'
```
meooow25: Just `return` would be fine, not necessary to assign it to `line`.

tle/cogs/codeforces.py
```diff
@@ -377,7 +378,52 @@ def strfilt(s):
...
+                        len(subs_by_contest_id[contest.id]),
+                        len(cf_common.cache2.problemset_cache.get_problemset(contest.id))))
+            except ProblemsetNotCached :
+                pass
+
+        full_solve = sorted(packed_contest_subs_problemset, key=lambda contest: (contest[2] - contest[1]))
+        full_solve = [(contest_id, solved, total) for contest_id, solved, total in full_solve if (total - solved) > 0]
+
+        if not full_solve:
+            await ctx.send('User has no contests to fullsolve')
```
meooow25: Should ideally be raising a cog error here but since this cog does not have one, you can add it or leave it as is, up to you.

tle/cogs/codeforces.py
```diff
@@ -377,7 +378,52 @@ def strfilt(s):
...
+                line = f'[{contest.name}]({contest.url})\N{EN SPACE}[{solved}/{total}]'
+                return line
+
+            def make_page(chunk):
+                message = f'fullsolve list for `{member.display_name}`'
+                log_str = '\n'.join(make_line(entry) for entry in chunk)
+                embed = discord_common.cf_color_embed(description=log_str)
+                return message, embed
+
+            member = ctx.author
```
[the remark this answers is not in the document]
kabeer27: I was using member for member.display_name which i have now changed to ctx.author.display_name, should have declared member inside the member now i realise there is not much use for the variable

tle/cogs/codeforces.py
```diff
@@ -377,7 +378,52 @@ def strfilt(s):
...
+                if len(subs_by_contest_id[contest.id]) > 0:
+                    packed_contest_subs_problemset.append((
+                        cf_common.cache2.contest_cache.get_contest(contest.id),
+                        len(subs_by_contest_id[contest.id]),
+                        len(cf_common.cache2.problemset_cache.get_problemset(contest.id))))
+            except ProblemsetNotCached :
+                pass
+
+        full_solve = sorted(packed_contest_subs_problemset, key=lambda contest: (contest[2] - contest[1]))
+        full_solve = [(contest_id, solved, total) for contest_id, solved, total in full_solve if (total - solved) > 0]
```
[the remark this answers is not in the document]
kabeer27: I will be also checking if num_solved > 0 as else it would unnecessarily add around 150 (on average) pages of 0 submission contests.

tle/cogs/codeforces.py
kabeer27: I'll leave it as it is for now

tle/cogs/codeforces.py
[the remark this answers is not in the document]
meooow25: Whoops, missed that. That's needed too, yes.

tle/cogs/codeforces.py
```diff
@@ -377,7 +380,51 @@ def strfilt(s):
...
+                if num_solved>0 and num_solved < num_problems:
+                    contest_unsolved_pairs.append((contest, num_problems - num_solved))
+            except cache_system2.ProblemsetNotCached :
+                """In case of recent contents or cetain bugged contests"""
+                pass
+
+        contest_unsolved_pairs.sort(key = lambda p: (p[1], -p[0].startTimeSeconds))
+
+        if not contest_unsolved_pairs:
+            await ctx.send('User has no contests to fullsolve')
```
algmyr: I prefer doing early exit to having the if else.
```
# so like this
if x:
    do stuff
    return
do other stuff

# rather than
if x:
    do stuff
else:
    do other stuff
```
But it's mostly personal preference, doing that in general tends to keep down the indent level.

PR #203, opened 2019-10-28
Automated rank role updates

### LINKED ISSUE #154: Show role changes for updateroles command
It will be nice to show role changes for every user on `;updateroles`

### LINKED ISSUE #164: Automate update roles
We have a rating change monitor and an event system. Put them together and :chart_with_upwards_trend:

meooow25:
## Automated rank updates 🎉 

Resolves #154.
Resolves #164.

### Changes
- Added event listener support.
- Added listener to publish rank updates and top 5 rating increases.
- Added support for manual publishing.

### Screenshot
image

### DIFF
--- a/tle/cogs/handles.py
+++ b/tle/cogs/handles.py
@@ -6,9 +6,11 @@ import discord
 import random
 from discord.ext import commands
 
+from tle.util import cache_system2
 from tle.util import codeforces_api as cf
 from tle.util import codeforces_common as cf_common
 from tle.util import discord_common
+from tle.util import events
 from tle.util import paginator
 from tle.util import table
 from tle import constants
@@ -18,6 +20,7 @@ from PIL import Image, ImageFont, ImageDraw
 _HANDLES_PER_PAGE = 15
 _NAME_MAX_LEN = 20
 _PAGINATE_WAIT_TIME = 5 * 60  # 5 minutes
+_TOP_DELTAS_SHOW_COUNT = 5
 
 
 class HandleCogError(commands.CommandError):
@@ -147,6 +150,33 @@ class Handles(commands.Cog):
         self.bot = bot
         self.logger = logging.getLogger(self.__class__.__name__)
         self.font = ImageFont.truetype(constants.NOTO_SANS_CJK_BOLD_FONT_PATH, size=26) # font for ;handle pretty
+
+    @commands.Cog.listener()
+    async def on_ready(self):
+        cf_common.event_sys.add_listener(self._on_rating_changes)
+
+    @events.listener_spec(name='RatingChangesListener',
+                          event_cls=events.RatingChangesUpdate,
+                          with_lock=True)
+    async def _on_rating_changes(self, event):
+        contest, changes = event.contest, event.rating_changes
+        change_by_handle = {change.handle: change for change in changes}
+
+        async def rank_update_for_guild(guild):
+            channel_id = cf_common.user_db.get_rankup_channel(guild.id)
+            channel = guild.get_channel(channel_id)
+            if channel is None:
+                return
+            try:
+                await self._update_ranks(guild)
+                embed = self._make_rankup_embed(guild, contest, change_by_handle)
+                await channel.send(embed=embed.set_footer(text='Roles updated!'))
+            except HandleCogError:
+                pass
+
+        await asyncio.gather(*(rank_update_for_guild(guild) for guild in self.bot.guilds),
+                             return_exceptions=True)
+        self.logger.info(f'Roles updated for all guilds for contest {contest.id}.')
 
     @commands.group(brief='Commands that have to do with handles', invoke_without_command=True)
     async def handle(self, ctx):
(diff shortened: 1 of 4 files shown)


meooow25:
**Note on listeners:**

- Why `Listeners` when we have `Tasks`?
The `wait_for` on an event will only trigger the first time if an event is fired, the task waits on its `func`, and meanwhile the event is fired again. This can be remedied by forking off the task's `func` using `asyncio.create_task` and not waiting on it.
However this doesn't help when the event is dispatched multiple times with no async context switch in between. In this PR this happens when multiple contests have their rating change update detected at once. One could introduce an `asyncio.wait(0)` between dispatching events, but that is hacky. An alternate would be to batch these together in the event, but that too feels hacky and moreover is specific to this situation.
Probably the `Waiter.for_event` option should be removed from tasks and its usage replaced by listeners.

- What's the `with_lock` parameter for?
Prevents multiple instances of the callback executing simultaneously. In this PR, if we make the (reasonable?) assumption that the listener triggered first will have its callback executed first, then we can publish the rank updates in the order of contest start times.

tle/util/cache_system2.py
```diff
@@ -418,24 +418,30 @@ def is_newly_finished_without_rating_changes(self, contest):
             return
 
         all_changes = await self._fetch(self.monitored_contests)
+        all_changes.sort(key=lambda pair: pair[1][0].ratingUpdateTimeSeconds)
```
algmyr: Random indexing is no fun for trying to understand the code. Sadly lambdas does not support unpacking. :(

tle/cogs/handles.py
```diff
@@ -18,6 +20,7 @@
 _HANDLES_PER_PAGE = 15
 _NAME_MAX_LEN = 20
 _PAGINATE_WAIT_TIME = 5 * 60  # 5 minutes
+_TOP_DELTAS_SHOW_COUNT = 5
```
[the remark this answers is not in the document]

tle/cogs/handles.py
```diff
@@ -148,6 +151,33 @@ def __init__(self, bot):
...
 
+    @commands.Cog.listener()
+    async def on_ready(self):
+        cf_common.event_sys.add_listener(self._on_rating_changes)
+
+    @events.listener_spec(name='RatingChangesListener',
+                          event_cls=events.RatingChangesUpdate,
+                          with_lock=True)
+    async def _on_rating_changes(self, event):
+        contest, changes = event.contest, event.rating_changes
```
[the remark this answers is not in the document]
meooow25: Agree that `contest, change = x, y` is not nice when x, y are random, however here I think it's fine.

tle/cogs/handles.py
```diff
@@ -315,12 +345,12 @@ def __init__(self, bot):
...
-        """Update Codeforces rank roles for everyone."""
-        res = cf_common.user_db.getallhandles()
-        member_handles = [(ctx.guild.get_member(int(user_id)), handle) for user_id, handle in res]
+    async def _update_ranks(self, guild):
+        """For each member in the guild, fetches their current ratings and updates their role if
+        required.
+        """
+        res = cf_common.user_db.get_handles_for_guild(guild.id)
+        member_handles = [(guild.get_member(int(user_id)), handle) for user_id, handle in res]
         member_handles = [(member, handle) for member, handle in member_handles if member is not None]
```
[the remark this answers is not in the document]
meooow25: :=

tle/util/cache_system2.py
meooow25: Agree this is not nice. Changed the variable name and added a comment.

tle/util/events.py
```diff
@@ -24,8 +61,105 @@ def __init__(self):
...
+                    await self.func(event)
+            else:
+                await self.func(event)
+        except asyncio.CancelledError:
+            raise
+        except:
+            self.logger.exception(f'Exception in listener `{self.name}`.')
+
+    def __eq__(self, other):
+        if other is None or not isinstance(other, Listener):
```
krofna: > if other is None

Is this not redundant?

tle/util/events.py
```diff
@@ -24,8 +61,105 @@ def __init__(self):
...
+            return False
+        return (self.event_cls, self.func) == (other.event_cls, other.func)
+
+    def __hash__(self):
```
krofna: is this used anywhere?

tle/cogs/handles.py
```diff
@@ -342,7 +372,141 @@ def __init__(self, bot):
...
+
+        def rating_to_displayable_rank(rating):
+            rank = cf.rating2rank(rating).title
+            role = rank_to_role.get(rank)
+            return role.mention if role else rank
+
+        rank_changes_str = []
+        for member, change in member_change_pairs:
+            cache = cf_common.cache2.rating_changes_cache
+            if (change.oldRating == 1500
```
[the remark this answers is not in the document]
meooow25: This is to avoid a db fetch unless necessary.

tle/cogs/handles.py
```diff
@@ -342,7 +372,141 @@ def __init__(self, bot):
...
+        for member, change in member_change_pairs[:_TOP_DELTAS_COUNT]:
+            delta = change.newRating - change.oldRating
+            if delta <= 0:
+                break
+            increase_str = (f'{member.mention} (`{change.handle}`): {change.oldRating} '
+                            f'\N{HORIZONTAL BAR} **{delta:+}** \N{LONG RIGHTWARDS ARROW} '
+                            f'{change.newRating}')
+            top_increases_str.append(increase_str)
+
+        desc = '\n'.join(rank_changes_str) or 'No rank changes'
```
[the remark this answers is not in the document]
meooow25: I did a rough calculation, it's fine. Description limit is 2000 chars and it seems very unlikely more than 10 people in the server will change ranks. 200 chars per row should be enough.

tle/util/events.py
meooow25: Yep, we need this because EventSystem will put Listeners in a set.

tle/util/events.py
algmyr: It is.

meooow25: You're right, removed.

PR #208, opened 2019-11-02
Allow admin to invalidate ongoing duel

krofna:


tle/cogs/duel.py
```diff
@@ -378,6 +378,22 @@ def make_page(chunk):
...
+    @duel.command(brief='Invalidate a duel')
+    @commands.has_role('Admin')
+    async def invalidate(self, ctx, member: discord.Member):
+        active = cf_common.user_db.check_duel_complete(member.id)
+        if not active:
+            raise DuelCogError(f'{member.display_name} is not in a duel.')
+
+        duelid, challenger_id, challengee_id, _, _, _, _ = active
+        challenger = ctx.guild.get_member(challenger_id)
+        challengee = ctx.guild.get_member(challengee_id)
```
algmyr: This could be moved down to just before the usage to be more lazy, but doesn't matter much.

PR #220, opened 2019-11-09
Stop catching errors caused by cog load failure.

### LINKED ISSUE #219: Do not start the bot when a cog fails to load

meooow25:
Closes #219


meooow25:
Seems like I considered catching these errors a feature when I made my first commit 👀

PR #231, opened 2019-11-29
Extend the invalidation time window

kuroni:
From 30 seconds to 1 minute, because people don't have inhuman reading speed.


cheran-senthil:
lgtm

PR #235, opened 2019-12-07
Update ;help duel invalidate

krofna:


cheran-senthil:
This is precisely why PRs are stupid, and we should push straight to master.

Issue #230, opened 2019-11-26
;fullsolve should also take ;vc argument like +div1


krofna:

krofna:
fixed in #236

PR #236, opened 2019-12-07
Allow tags in fullsolve

krofna:


krofna:
Closes #230

tle/cogs/codeforces.py
```diff
@@ -395,10 +395,17 @@ def strfilt(s):
             await ctx.send(f'Recommended contest for `{str_handles}`', embed=embed)
 
     @commands.command(brief="Display unsolved rounds closest to completion")
-    async def fullsolve(self, ctx):
-        """Displays a list of contests, sorted by number of unsolved problems"""
+    async def fullsolve(self, ctx, *args: str):
+        """Displays a list of contests, sorted by number of unsolved problems
+        e.g ;fullsolve +edu"""
```
meooow25: Add `usage = [keywords..]`, and maybe put in help that contest names matching any of the tags will be considered.
Maybe we should start accepting regex for these 🙃

krofna: accepting regex is kinda waturr

PR #237, opened 2019-12-08
Fix #233

### LINKED ISSUE #233: plot solved should have argument to plot contest-only or some combination
requested by ngk

krofna:


tle/util/codeforces_common.py
```diff
@@ -205,3 +205,24 @@ def days_ago(t):
...
+        types.append('OUT_OF_COMPETITION')
+        args.remove('+outof')
+    if '+virtual' in args:
+        types.append('VIRTUAL')
+        args.remove('+virtual')
+    if '+practice' in args:
+        types.append('PRACTICE')
+        args.remove('+practice')
+    types = types or ['CONTESTANT', 'OUT_OF_COMPETITION', 'VIRTUAL', 'PRACTICE']
+    return team, types
```
meooow25: I would prefer if this function did not modify `args` and instead returned a new list of leftovers after removing extracting its required values. Could return it as `team, types, leftovers`.
Also callers would not need to do `args = list(args)`.

tle/cogs/graphs.py
```diff
@@ -320,7 +321,10 @@ def __init__(self, bot):
 
     @plot.command(brief='Show histogram of solved problems on CF.')
     async def solved(self, ctx, *args: str):
-        """Shows a histogram of problems solved on Codeforces for the handles provided."""
+        """Shows a histogram of problems solved on Codeforces for the handles provided.
+        e.g. ;plot solved meooow +contest +virtual +outof +dp"""
```
meooow25: For both commands affected by it, it should be clear that (`+contest`, `+virtual`, `+outof`, `+practice`) are the submission type options since they are our chosen params. And rest would be accepted tags.

PR #241, opened 2019-12-13
Improve days_ago

krofna:


meooow25:
What's this for?

krofna:
Several users pointed out that it's non-intuitive that it never says 2 days ago, and yesterday was 1 day ago and not 2 days ago.

meooow25:
Got it. Then the only change you need is `math.ceil` -> `math.floor`.
And maybe `<=` -> `<`.

PR #252, opened 2019-12-17
Dont recommend contests with problems solved on other division

### LINKED ISSUE #116: ;vc sometimes recommends contests with solved problems
image

krofna:
Closes #116

### DIFF
--- a/tle/cogs/codeforces.py
+++ b/tle/cogs/codeforces.py
@@ -363,13 +363,14 @@ class Codeforces(commands.Cog):
         else:
             divs = [strfilt(x) for x in markers]
 
+        contest_id_map = {contest.id: contest for contest in contests}
+        bad_times = {contest_id_map.get(sub.problem.contestId).startTimeSeconds for problem in user_submissions
+                     if sub.problem.contestId and sub.problem.contestId < cf.GYM_ID_THRESHOLD}
+
         recommendations = {contest.id for contest in contests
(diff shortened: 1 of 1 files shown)


tle/cogs/codeforces.py
```diff
@@ -363,13 +363,14 @@ def strfilt(s):
         else:
             divs = [strfilt(x) for x in markers]
 
+        contest_id_map = {contest.id: contest for contest in contests}
```
meooow25: Directly map to start times? That would simplify the next statement to
```py
bad_times = {contest_id_to_start_map.get(sub.problem.contestId)
             for subs in user_submissions for sub in subs}
```
The set might end up with `None` but it won't harm anything.

tle/cogs/codeforces.py
```diff
@@ -363,13 +363,14 @@ def strfilt(s):
...
+        bad_times = {contest_id_map.get(sub.problem.contestId).startTimeSeconds for subs in user_submissions
+                     for sub in subs if sub.problem.contestId and sub.problem.contestId < cf.GYM_ID_THRESHOLD}
```
meooow25: Could you add a short comment saying we are using contest start time as we have no better way and this might exclude some problems (like if someone solved div2A, then div1 is still completely unsolved), because we don't care, I guess?

krofna:
fixed differently

PR #254, opened 2019-12-22
Split logic into filters.

meooow25:
Try to clean code.
Remove rating filter from ;plot solved because it doesn't make sense.
Remove rating and type filter from ;plot scatter because it doesn't make sense.
I am sleepy and there may be bugs.

### DIFF
--- a/tle/cogs/graphs.py
+++ b/tle/cogs/graphs.py
@@ -20,6 +20,7 @@ from tle import constants
 from tle.util import codeforces_api as cf
 from tle.util import codeforces_common as cf_common
 from tle.util import discord_common
+from tle.util import filters
 
 pd.plotting.register_matplotlib_converters()
 
@@ -30,12 +31,6 @@ CONTEST_ACTIVE_TIME_CUTOFF = 90 * 24 * 60 * 60 # 90 days
 class GraphCogError(commands.CommandError):
     pass
 
-def nice_sub_type(types):
-    nice_map = {'CONTESTANT':'Contest: {}',
-                'OUT_OF_COMPETITION':'Unofficial: {}',
-                'VIRTUAL':'Virtual: {}',
-                'PRACTICE':'Practice: {}'}
-    return [nice_map[t] for t in types]
 
 def _get_current_figure_as_file():
     filename = os.path.join(constants.TEMP_DIR, f'tempplot_{time.time()}.png')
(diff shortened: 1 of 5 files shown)


tle/util/filters.py
```diff
@@ -0,0 +1,186 @@
...
+        problem = submission.problem
+        if problem.rating is None:
+            return self.missing_ok
+        return self.low <= problem.rating <= self.high
+
+
+class TimeFilter:
+    def __init__(self, *, low=None, high=None):
+        self.low = low if low is not None else 0
+        self.high = high if high is not None else 10 ** 10
```
algmyr: This limit bothers me for no real reason. It's large enough to not be a problem (250 years in the future), but small enough to still be a tangible time since epoch. Nothing to change, just something I noted.

tle/util/filters.py
```diff
@@ -0,0 +1,186 @@
...
+                fmt = '%d%m%Y'
+            elif len(arg) == 6:
+                fmt = '%m%Y'
+            elif len(arg) == 4:
+                fmt = '%Y'
+            else:
+                raise ValueError
+            return dt.datetime.strptime(arg, fmt).timestamp()
+        except ValueError:
+            raise ParamParseError(f'`{arg}` is an invalid date argument.')
```
algmyr: I guess catching everything thrown by the strptime and raising a more generic error is intended? If you would want to keep the more specific ValueErrors thrown by strptime you could move the string to when the ValueError is raised, and then re-raise the new error with the same error message.

tle/cogs/graphs.py
```diff
@@ -306,41 +309,34 @@ def __init__(self, bot):
     async def solved(self, ctx, *args: str):
         """Shows a histogram of problems solved on Codeforces for the handles provided.
         e.g. ;plot solved meooow +contest +virtual +outof +dp"""
-        team, _, types, tags, dlo, dhi, rlo, rhi, args = cf_common.filter_sub_args(args)
+        predicates, args = filters.parse_all_from(
+            args, filters.TeamFilter, filters.TimeFilter)
+        types_pred, args = filters.TypeFilter.parse_from(args)
+        tags_pred, args = filters.TagFilter.parse_from(args)
```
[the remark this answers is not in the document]
meooow25: Because `types_pred` is required later and `tags_pred` must be last one to be parsed.
`tags_pred` would take everything starting with `+` as a tag, including `+contest` for example. I would prefer a more command-line like format (`--tags=dp,string`), but @krofna is not in favor.

tle/util/filters.py
meooow25: It was intentional to hide the `ValueError` from `strptime`. Its error messages look like `"time data 'inputstring' does not match format '%m%Y'"` which we wouldn't want to go to the user.
I should add `from originalError` though.

tle/util/filters.py
meooow25: Could replace it with `float('inf')`? 🤷‍♂

meooow25:
Obsolete

PR #279, opened 2020-02-17
Dogs rise up.

aryanc403:
Make sure that only people with this role can ping it.


krofna:
useless

aryanc403:
If role in the server has a purpose then so does this PR for dog community in server.

krofna:
Then we might as well add `;role give Purgatory` if people want to ban themselves from #general

PR #282, opened 2020-02-23
Implement ;duel rating plot

krofna:


tle/cogs/duel.py
```diff
@@ -5,21 +5,53 @@
...
+from matplotlib import pyplot as plt
 
 from tle.util.db.user_db_conn import Duel, Winner
 from tle.util import codeforces_api as cf
 from tle.util import codeforces_common as cf_common
 from tle.util import paginator
 from tle.util import discord_common
 from tle.util import table
-
+from tle.cogs.graphs import _get_current_figure_as_file
```
algmyr: cogs depending on each other seems like a bad idea. It might make sense to split out some plotting logic in a utility used by both.

tle/cogs/duel.py
```diff
@@ -5,21 +5,53 @@
...
+    DuelRank(-10 ** 9, 1300, 'Newbie', 'N', '#CCCCCC', 0x808080),
+    DuelRank(1300, 1400, 'Pupil', 'P', '#77FF77', 0x008000),
+    DuelRank(1400, 1500, 'Specialist', 'S', '#77DDBB', 0x03a89e),
+    DuelRank(1500, 1600, 'Expert', 'E', '#AAAAFF', 0x0000ff),
+    DuelRank(1600, 1700, 'Candidate Master', 'CM', '#FF88FF', 0xaa00aa),
+    DuelRank(1700, 1800, 'Master', 'M', '#FFCC88', 0xff8c00),
+    DuelRank(1800, 1900, 'International Master', 'IM', '#FFBB55', 0xf57500),
+    DuelRank(1900, 2000, 'Grandmaster', 'GM', '#FF7777', 0xff3030),
+    DuelRank(2000, 2100, 'International Grandmaster', 'IGM', '#FF3333', 0xff0000),
+    DuelRank(2100, 10 ** 9, 'Legendary Grandmaster', 'LGM', '#AA0000', 0xcc0000)
```
algmyr: Not specific to this code, but sometime we should really tidy up so that we don't use two different formats for colors. It's been an eyesore since the beginning.

tle/cogs/duel.py
```diff
@@ -5,21 +5,53 @@
...
+            return rank
+
+def _plot_rating_bg():
+    bgcolor = plt.gca().get_facecolor()
+    for rank in DUEL_RANKS:
+        plt.axhspan(rank.low, rank.high, facecolor=rank.color_graph, alpha=0.8, edgecolor=bgcolor, linewidth=0.5)
+
+    locs, labels = plt.xticks()
+    for loc in locs:
+        plt.axvline(loc, color=bgcolor, linewidth=0.5)
```
algmyr: This is duplicated code, right? Might make sense to split out to some common places. Especially the second one.

tle/cogs/duel.py
```diff
@@ -443,6 +475,70 @@ def make_page(chunk, page_num):
...
+            else:
+                delta = round(elo_delta(challenger_r, challengee_r, 0.5))
+
+            rating[challenger] = challenger_r + delta
+            rating[challengee] = challengee_r - delta
+            if challenger in duelists or challengee in duelists:
+                if challenger in duelists:
+                    plot_data[challenger].append((time_tick, rating[challenger]))
+                if challengee in duelists:
+                    plot_data[challengee].append((time_tick, rating[challengee]))
```
algmyr: Is `time_tick` the right thing to plot on x? Shouldn't we use something like finish time?Although maybe that gets weird since people do duels in spurts.

tle/cogs/duel.py
```diff
@@ -443,6 +475,70 @@ def make_page(chunk, page_num):
...
+                time_tick += 1
+
+        if time_tick == 0:
+            raise DuelCogError(f'Nothing to plot.')
+
+        plt.clf()
+        min_rating = 1350
+        max_rating = 1550
```
algmyr: Why these values? If there is a good reason a comment would be nice.

tle/cogs/duel.py
```diff
@@ -443,6 +475,70 @@ def make_page(chunk, page_num):
...
+        for duelist, rating_data in plot_data.items():
```
algmyr: NIT: Could do values() here since we don't care about the key.

tle/cogs/duel.py
```diff
@@ -443,6 +475,70 @@ def make_page(chunk, page_num):
...
+                     marker='o',
+                     markersize=2,
+                     markerfacecolor='white',
+                     markeredgewidth=0.5)
+
+        _plot_rating_bg()
+        plt.xlim(0, time_tick - 1)
+        plt.ylim(min_rating - 100, max_rating + 100)
+
+        labels = [f'{ctx.guild.get_member(duelist).display_name} ({rating_data[-1][1]})'
```
algmyr: NIT: This is a case where .format would be nicer. Easier to glance what the format actually is.

tle/cogs/duel.py
krofna: Originally i plotted date like in `;plot rating` and it looked terrible

tle/cogs/duel.py
krofna: i wanted visible color span to be the same as with the cf rating plot.

Issue #281, opened 2020-02-22
Don't allow vjudge accounts (and similar accs) to register


algmyr:
This restriction is already implemented in the plotting commands to avoid fetching tens of thousands of pages of submissions. We should add this restriction to `;handle identify` also.

algmyr:
Fixed by #287

PR #287, opened 2020-03-06
Prevent identifying as VJudge handles

kuroni:
Resolving #281


krofna:
This looks wrong. You should instead raise HandleIsVjudgeError and make handle cog catch it.

PR #289, opened 2020-03-09
Switch gudgitters to an image

krofna:


tle/cogs/handles.py
```diff
@@ -60,6 +68,104 @@ def rating_to_color(rating):
...
+    layout.set_ellipsize(Pango.EllipsizeMode.END)
+
+    i = 0
+    BG1 = (0.95, 0.95, 0.95)
+    BG2 = (0.9, 0.9, 0.9)
+
+    def draw_bg(y):
+        nonlocal i
+        i += 1
+        p = i%2
```
algmyr: I think this was some hacky old code. Let's remove internal state from the function and pass some odd/even variable as an argument instead.

tle/cogs/handles.py
```diff
@@ -60,6 +68,104 @@ def rating_to_color(rating):
...
+    context.set_source_rgb(*DISCORD_GRAY)
+    context.rectangle(0, 0, WIDTH, HEIGHT)
+    context.fill()
+    layout = PangoCairo.create_layout(context)
+    layout.set_font_description(Pango.font_description_from_string(','.join(FONTS) + ' 20'))
+    layout.set_ellipsize(Pango.EllipsizeMode.END)
+
+    i = 0
+    BG1 = (0.95, 0.95, 0.95)
+    BG2 = (0.9, 0.9, 0.9)
```
algmyr: Move this up to the other colors?

tle/cogs/handles.py
```diff
@@ -60,6 +68,104 @@ def rating_to_color(rating):
...
+        # Simple
+        context.move_to(BORDER_MARGIN, y)
+        context.line_to(WIDTH, y)
+        context.line_to(WIDTH, nxty)
+        context.line_to(0, nxty)
+
+        if p:
+            context.set_source_rgb(*BG1)
+        else:
+            context.set_source_rgb(*BG2)
```
algmyr: Could have a list of two colors like `ROW_COLORS` and index by the parity.

tle/cogs/handles.py
```diff
@@ -60,6 +68,104 @@ def rating_to_color(rating):
...
+
+        context.fill()
+
+    def draw_row(pos, username, handle, rating, color, y, bold=False):
+        context.set_source_rgb(*[x/256.0 for x in color])
```
algmyr: That should be 255.0

tle/cogs/handles.py
```diff
@@ -60,6 +68,104 @@ def rating_to_color(rating):
...
+        draw_row(str(pos), name, handle, str(rating), color, y)
+        if rating != 'N/A' and rating >= 3000:  # nutella
+            draw_row('', name[0], handle[0], '', BLACK, y)
+        y += LINE_HEIGHT
+
+    filename = os.path.join(constants.TEMP_DIR, f'tempgitters_{time.time()}.png')
+    surface.write_to_png(filename)
+    with open(filename, 'rb') as file:
+        discord_file = discord.File(io.BytesIO(file.read()), filename='gudgitters.png')
+    os.remove(filename)
```
algmyr: I don't think this should be needed, there ought to be a way to write to BytesIO object (or similar) directly.

tle/cogs/handles.py
```diff
@@ -60,6 +68,104 @@ def rating_to_color(rating):
...
+            layout.set_width((width - COLUMN_MARGIN)*1000) # pixel = 1000 pango units
+            layout.set_markup(text, -1)
+            PangoCairo.show_layout(context, layout)
+            context.rel_move_to(width, 0)
+
+        draw(pos, WIDTH_RANK)
+        draw(username, WIDTH_NAME)
+        draw(handle, WIDTH_NAME)
+        draw(rating)
+
```
algmyr: I would maybe make some separator here just to signal the end up helper functions and the start of the actual logic

tle/cogs/handles.py
```diff
@@ -60,6 +68,104 @@ def rating_to_color(rating):
...
+        nonlocal i
+        i += 1
+        p = i%2
+        nxty = y + LINE_HEIGHT
+
+        # Simple
+        context.move_to(BORDER_MARGIN, y)
+        context.line_to(WIDTH, y)
+        context.line_to(WIDTH, nxty)
+        context.line_to(0, nxty)
```
algmyr: If we've settled on rectangle, let's use https://cairographics.org/manual/cairo-Paths.html#cairo-rectangle to be more self documenting.

tle/cogs/handles.py
```diff
@@ -104,31 +106,19 @@ def get_gudgitters_image(rankings):
...
 
-    i = 0
-    BG1 = (0.95, 0.95, 0.95)
-    BG2 = (0.9, 0.9, 0.9)
-
-    def draw_bg(y):
-        nonlocal i
-        i += 1
-        p = i%2
+    def draw_bg(y, p):
```
algmyr, line 109: Maybe have a more descriptive name than `p`

PR #299, opened 2020-04-11
Replace all uses of zero width space and \0

algmyr:
Only actual strings are checked for leading underscores, and in the end `str` is called on the object passes. So passing a wrapped string will bypass the checks.

### DIFF
--- a/tle/cogs/graphs.py
+++ b/tle/cogs/graphs.py
@@ -224,7 +224,7 @@ class Graphs(commands.Cog):
         plt.clf()
         _plot_rating(resp)
         current_ratings = [rating_changes[-1].newRating if rating_changes else 'Unrated' for rating_changes in resp]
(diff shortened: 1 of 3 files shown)


tle/cogs/graphs.py
```diff
@@ -371,7 +371,7 @@ def __init__(self, bot):
             # NOTE: matplotlib ignores labels that begin with _
             # https://matplotlib.org/api/pyplot_api.html#matplotlib.pyplot.legend
             # Add zero-width space to work around this
```
meooow25, line 373: Probably this explanation can be moved to the `StrWrap` class.

algmyr:
> This is pretty hacky but current solution isn't great either, so ¯\_(ツ)_/¯

This is hacky, but avoids some font issues that showed up with zero width spaces/null chars. I wouldn't have bothered to dig into the matplotlib code otherwise. :)

Issue #224, opened 2019-11-16
Ranklist prediction improvement


meooow25:
- Calculate rating of a person from last rating change instead of sum of deltas to make it more robust.
- Maybe add a command to randomly sample a running contest standings to make sure current rating values are ok to use for prediction.

krofna:
instead query user.info for ratings

krofna:
fixed in #301

PR #301, opened 2020-04-11
Query ratedList to get the user ratings for predicting rating changes instead of using userInfo and its cache.

Mohammad-Yasser:
We need to avoid using the cache of the userInfo while predicting changes because it has some problems.
Querying `user.info` with all the handles is impossible to be done in one request because the URL will be too large and so we'll need to send multiple requests.
So we agreed on querying `ratedList` instead.

### DIFF
--- a/tle/util/cache_system2.py
+++ b/tle/util/cache_system2.py
@@ -592,9 +592,7 @@ class RanklistCache:
                 # The contest is not rated
                 ranklist = Ranklist(contest, problems, standings, now, is_rated=False)
             else:
-                ratedList = await CacheSystem.ratedList(activeOnly=False)
-                current_rating = {user.handle: user.effective_rating
-                                  for user in ratedList}
+                current_rating = CacheSystem.getUsersEffectiveRating(activeOnly=False)
                 if 'Educational' in contest.name:
                     # For some reason educational contests return all contestants in ranklist even
                     # when unofficial contestants are not requested.
@@ -636,5 +634,10 @@ class CacheSystem:
 
     @ttl_cache(ttl = 30 * 60)
     @staticmethod
-    async def ratedList(*, activeOnly=None):
-        return cf.user.ratedList(activeOnly=activeOnly)
+    async def getUsersEffectiveRating(*, activeOnly=None):
+        """ Returns a dictionary mapping user handle to his effective rating for all the users.
+        """
+        ratedList = await cf.user.ratedList(activeOnly=activeOnly)
+        users_effective_rating_dict = {user.handle: user.effective_rating
+                                  for user in ratedList}
+        return users_effective_rating_dict
--- a/tle/util/cache_system2.py
+++ b/tle/util/cache_system2.py
@@ -591,7 +591,7 @@ class RanklistCache:
                 # The contest is not rated
                 ranklist = Ranklist(contest, problems, standings, now, is_rated=False)
             else:
-                ratedList = await cf.user.ratedList()
+                ratedList = await cf.user.ratedList(activeOnly=False)
                 current_rating = {user.handle: user.effective_rating
                                   for user in ratedList}
                 if 'Educational' in contest.name:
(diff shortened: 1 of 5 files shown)


krofna:
Closes #224

tle/util/cache_system2.py
```diff
@@ -592,10 +591,9 @@ async def generate_ranklist(self, contest_id, *, fetch_changes=False, predict_ch
                 # The contest is not rated
                 ranklist = Ranklist(contest, problems, standings, now, is_rated=False)
             else:
-                get = self.cache_master.rating_changes_cache.get_current_rating
-                current_rating = {row.party.members[0].handle: get(row.party.members[0].handle,
-                                                                   default_if_absent=True)
-                                  for row in standings_official}
+                ratedList = await cf.user.ratedList()
```
meooow25, line 594: It should be `cf.user.ratedList(active_only=False)` or you will only get active users.

Mohammad-Yasser:
I used `ttl_cache` for `user.ratedList()`. Didn't test it though, as it needs a running contest. I know I can fake a running contest in the code, but don't find it worthy, as I'm just using a very simple decorator. And any way we can wait for tomorrow's contest and test it.

Mohammad-Yasser:
1. Do you mean to implement the decorator by myself? But why do this if there is a package already doing it?
2. Why? We want to provide this cache for anyone calling this API method, right?

meooow25:
1. You do not have to implement a decorator. It is as simple as keeping the ratings and the last fetch time somewhere, so that you can refetch if it is too old.
2. No, we definitely don't want every caller to get a cached response. When someone calls any function in `codeforces_api` they expect it to retrieve the data from the API.

Mohammad-Yasser:
1. Hmm .. I don't think that this is a best practice.
Generally, if there are parameters, then I'll have to create a dictionary mapping each possible assignment for the parameters to the cached value. In this case, we only need one instance yes. But this will be something that will only be used for this case, and if something changes like we need to cache also (`isActive=true`) or introduce a new parameter then it'll be a mess. And we may need in the future to have some cache for other functions too, or so.
This way is tightly coupled to the current state of the function (which yes we don't expect to change) and our current needs (which we are not sure if they change or not).
The decorator is way more abstract, flexible and easily extensible. 

2. Oh yes. At first, I was thinking that this is something that changes mainly every couple of days at least, so it won't harm anyone if the user gets a ~20-minute old response. But this way we'll be restricting the user, and of course, he may need to get a fresh response.
But I think it should be placed in `CacheSystem` itself? 

I have a proposal. We can make a function in `CacheSystem`, that calls the `codeforces_api` method. That way, anyone who doesn't mind a cached version, will use the `CacheSystem` method, otherwise, he'll use the `codeforces_api` one. Also, this way is more abstract and extensible. What do you think?

meooow25:
1. If you expect TTL caches (or some other functionality offered by cachetools) may be used in other places too, then sure, it makes sense and I won't object.
2. That sounds reasonable, we can go with that 👍

tle/util/cache_system2.py
```diff
@@ -634,3 +633,8 @@ async def run(self):
         await self.contest_cache.run()
         await self.problem_cache.run()
         await self.problemset_cache.run()
+
+    @ttl_cache(ttl = 30 * 60)
+    @staticmethod
+    async def ratedList(*, activeOnly=None):
+        return cf.user.ratedList(activeOnly=activeOnly)
```
meooow25, line 640: You could cache the handle to rating dict, instead of all the user info, to reduce memory usage.
The total JSON data sent by the server is around 82MB at the moment, which is not negligible, and at the moment we only care about the ratings.

Mohammad-Yasser:
@meooow25 The other caching library isn't for async functions, so I used [this library](https://pypi.org/project/aiocache/) instead.
I tested the PR on the ongoing education round, and it works fine.

PR #304, opened 2020-04-14
Make deployments easy.

Mohammad-Yasser:
It may be needed to deploy on a remote machine to experiment with a specific branch, so instead of doing it manually, we can enable doing this via the bot itself.

There are two new cogs:
- `;git set_origin_uri origin_uri`: does `git remote set-url origin $origin_uri`.
- `;git checkout branch_name`: does `git fetch origin/$branch_name` followed by `git checkout origin/$branch_name`.

After checking out to the remote branch, you can use the `restart` cog to deploy the corresponding version.

Make sure that you store your credentials via this command:
`git config --global credential.helper store`


run.sh
```diff
@@ -6,8 +6,14 @@ cd "$(dirname "$0")"
 [ -e environment ] && . ./environment
 
 while true; do
-    git pull
-    FONTCONFIG_FILE=$PWD/extra/fonts.conf poetry run python -m tle
+    # Make sure the following command doesn't need credentials.
+    # You can store your credentials using: git config --global credential.helper store
+    git remote set-url origin "$ORIGIN_URI"
+    git fetch origin
+    git reset --hard "$COMMIT_HASH"
```
algmyr, line 13: I'm not sure I'm happy with doing this at every restart. A pull makes sense since it just brings stuff up to date. This is extra information and requires you to manually update things like the commit hash every time you update your tracking branch.

I might be more interesting to allow admins to call git commands (or some very restricted version of them) from the server. E.g. if we were to make `;git` a command group we could allow something like `;git set_origin <origin>` and `;git reset_to_commit <hash>`, maybe throw in `;git pull` as well and remove the call from the script. If you're feeling particularly adventurous you could even make `;git ...` just straight up call git with a bunch of arguments, but this is questionable at best from a security standpoint.

.gitignore
```diff
@@ -91,6 +91,12 @@ ENV/
 env.bak/
 venv.bak/
 
+# Env vars
+environment
+ORIGIN_URI
+BRANCH_NAME
+COMMIT_HASH
```
algmyr, line 98: Not really a fan of using files to store individual environment variables. See the idea of instead interacting with git through the bot.

run.sh
```diff
@@ -6,8 +6,14 @@ cd "$(dirname "$0")"
...
+
+    FONTCONFIG_FILE=$PWD/extra/fonts.conf
+    poetry run python -m tle
```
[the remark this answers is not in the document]
Mohammad-Yasser: Ah I see, I didn't know that there was a reason for making them in one line.

run.sh
Mohammad-Yasser: We could avoid doing it at every restart through having a different cog: `;meta deploy` where it exits using a different status code and then in the script handle this code to deploy.
But I agree that having `;git` command group looks nice. I also agree we don't need to support general `git` commands. We only need `git set_origin <origin>` and `git reset <hash>`.

However, I don't understand why the `git history` command is done that way in the code. Why isn't it just `os.system(..)`?

algmyr: Yes. `os.system` does not give any way of capturing stdout. Presenting the output of the command in a nice way is the whole point of the history command. I would recommend to keep using Popen for the other commands and reporting (in some way) the response of running the command (to indicate success/error/...). You could technically do this with `os.system` and looking at the exit code, but you will not get any meaningful errors other than "something went wrong".

Mohammad-Yasser:
I didn't test it yet, as it's just a draft. I agree that this way is nicer than the previous one.
Any further suggestions are much appreciated.

Mohammad-Yasser:
I found this package which may be nicer to use: https://github.com/gitpython-developers/GitPython/
What do you think?

Mohammad-Yasser:
For the previously mentioned `gitpython` library, it's written [here](https://gitpython.readthedocs.io/en/stable/intro.html#leakage-of-system-resources) that it's not suitable for daemons, so I guess it's not suitable for the bot.

I thought of a simpler way to deploy, without needing to do any local changes or disrupt the git history. Just fetch the remote branch, and checkout to it.

What do you think? @algmyr

algmyr:
Having set-origin, fetch and checkout sounds good to me.

tle/cogs/git.py
```diff
@@ -0,0 +1,97 @@
...
+
+    @commands.group(brief='Git commands', invoke_without_command=True)
+    async def git(self, ctx):
+        """Do git commands in the bot repo."""
+        await ctx.send_help(ctx.command)
+
+    @git.command(brief='git remote set-url origin $https_origin_uri', usage='[https_origin_uri]')
+    @commands.has_role('Admin')
+    async def set_origin_uri(self, ctx, origin_uri):
+        git_set_origin(origin_uri)
```
algmyr, line 81: Should these make use of the returned strings/errors? Right now they are just dropped?

Mohammad-Yasser: Well, I was also wondering about this.
I think that if the operation is successful, then the message may have no importance, right?
The other simple option is to show the whole git output in discord, which I think is okay too, but I don't think it's really needed, so I just preferred to show a success message.

algmyr: Right now it always print as if it was successful. Even if there is an error. Would it make sense to return the response string you want from the git_* commands? Also, for consistency name those methods with a leading _ to suggest they are "private" helper functions.

Mohammad-Yasser:
After tedious debugging, I found out that the output of `git fetch` and `git checkout` is sent through `stderr`, not `stdout`, so we have to capture that too. 

I tested the PR and it's working fine.

tle/cogs/git.py
```diff
@@ -0,0 +1,98 @@
...
+    @commands.group(brief='Git commands', invoke_without_command=True)
+    async def git(self, ctx):
+        """Do git commands in the bot repo."""
+        await ctx.send_help(ctx.command)
+
+    @git.command(brief='git remote set-url origin $https_origin_uri', usage='[https_origin_uri]')
+    @commands.has_role('Admin')
+    async def set_origin_uri(self, ctx, origin_uri):
+        self.logger.info(f'Setting origin uri to {origin_uri}')
+        _git_set_origin(origin_uri) # It doesn't have any output.
```
algmyr, line 78: It does have output if OSError or if origin_uri is invalid.

tle/cogs/git.py
```diff
@@ -0,0 +1,98 @@
...
+    env = {}
+    for k in ['SYSTEMROOT', 'PATH']:
+        v = os.environ.get(k)
+        if v is not None:
+            env[k] = v
+    # LANGUAGE is used on win32
+    env['LANGUAGE'] = 'C'
+    env['LANG'] = 'C'
+    env['LC_ALL'] = 'C'
+    out = subprocess.Popen(cmd, stdout = subprocess.PIPE, env=env, stderr = subprocess.STDOUT).communicate()[0]
```
algmyr, line 28: What makes most sense? Merging stdout and stderr as done here, or extract them separately and put it together somehow? You can get them separately by doing `stderr = subprocess.PIPE`. You get a tuple of (stdout, stderr) as output.

You could create something like this, maybe even leave out either section if empty:

    Output:
    ```
    stdout here
    ```
    Error:
    ```
    stderr here
    ```

Mohammad-Yasser: Hmm I thought that it's weird to have normal expected output in the error stream, and I do not understand why `git fetch` and `git checkout` report to `stderr` instead of `stdout` in the first place.

So generally yes I think we should not merge `stdout` and `stderr`, but the case of git, I cannot get why in the first place it's using `stderr`, and I think anyone will expect informative messages that don't report any problems will not be considered as errors. 

So if I do `;git fetch`, if I do as you propose, the user will be shown:
```
Error:
From https://github.com/Mohammad-Yasser/TLE
 * branch            easy-deployments -> FETCH_HEAD
```
and `;git checkout`, 
```
Error:
Previous HEAD position was 2aa4286 Query ratedList to get the user ratings for predicting rating changes instead of using userInfo and its cache. (#301)
HEAD is now at 06efbc0 Renaming private functions
```
Which I don't think makes sense, and maybe confusing for the users, because they are not errors at all.

Anyway, you (the admins) will be the ones using it, so you can decide what you like, and I'll do what I like in my other fork. :P

algmyr:
If you think yaml syntax highlighting looks good then I think this looks good.

PR #305, opened 2020-04-25
Split too large user.info requests

Mohammad-Yasser:
If `user.info`request has too many handles, the request URL will too large and thus the request will fail with [this error](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/414).

So, a simple fix is to split the too large request into small enough requests.

The possible drawback here is that the request will take significantly more time, due to rate-limiting.

### DIFF
--- a/tle/util/codeforces_api.py
+++ b/tle/util/codeforces_api.py
@@ -16,6 +16,7 @@ PROFILE_BASE_URL = 'https://codeforces.com/profile/'
 ACMSGURU_BASE_URL = 'https://codeforces.com/problemsets/acmsguru/'
 GYM_ID_THRESHOLD = 100000
 DEFAULT_RATING = 1500
+MAX_HANDLES_PER_QUERY = 300 # To avoid sending too large requests.
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


tle/util/codeforces_api.py
```diff
@@ -335,6 +337,12 @@ async def problems(*, tags=None, problemset_name=None):
 class user:
     @staticmethod
     async def info(*, handles):
+        if len(handles) > MAX_HANDLES_PER_QUERY:
+            chunks = chunkify(handles, MAX_HANDLES_PER_QUERY)
+            logger.warning(f'cf.info request with {len(handles)} handles, \
+            will be chunkified into {len(chunks)} requests.')
```
algmyr, line 343: This will put whitespace in the string. You can do two format strings and use the fact that strings literals are implicitly concatenated (or add two strings, or whatever). E.g.
```
f'cf.info request with {len(handles)} '
f'will be chunkified into {len(chunks)} requests.'
```
will concatenate into one string.

tle/util/codeforces_api.py
```diff
@@ -335,6 +337,12 @@ async def problems(*, tags=None, problemset_name=None):
...
+            for chunk in chunks:
+                await user.info(handles=chunk)
```
algmyr, line 345: Is the returned values of this just dropped? Also, it seems odd to me do do this recursively.

What about making the chunks in any case, if number of chunks is >1 then warn, then just loop over chunks and do the work for every chunk. Have a list you `+=` with `[make_from_dict(User, user_dict) for user_dict in resp]` in each iteration and return that list.

PR #307, opened 2020-04-26
A rudimentary version of CF Moving Average rating

stefdasca:
Since the plots are already there, I believe this can be added quite easily.


krofna:
suggestion not a PR

PR #310, opened 2020-05-01
Filtered only official contestants for predicting

Mohammad-Yasser:
This is a quick fix. I know I should do `dict.get(handle, 1500)`, but I don't even understand why some users aren't in `ratedList`.

I tested it quickly and it's working.


meooow25:
Fixes bug introduced in #301.

PR #311, opened 2020-05-04
Add rated VC feature

Mohammad-Yasser:
This can be considered as an MVP of the feature of rated VCs.
It's about letting people do a rated CF VC.
Wanted to check your opinions regarding it and regarding the future steps.
I'll add in a comment the current flow, and the future steps.

No worries of course if you think it's an unneeded feature, I'll be using it in my fork anyway.


Mohammad-Yasser:
Current flow:
1. The admin calls for a rated vc.
2. People express their desire to participate.
3. We do `;vc handles +pattern` to get the possible contests.
4. We do `;ratedvc contest_id duration step handles` and ask people to start the vc immediately. `step` is not a good name ofc.
5. Each `step` seconds a ranklist will be displayed with the predicted deltas (in the vc rating).
6. After `duration` seconds, a final ranklist will be displayed and the `vc_rating`s will be updated with the deltas.

tle/cogs/contests.py
```diff
@@ -461,10 +467,59 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+            handles.remove('+server')
+            guild_handles = [handle for discord_id, handle
+                             in cf_common.user_db.get_handles_for_guild(ctx.guild.id)]
+            handles.update(guild_handles)
+        handles = await cf_common.resolve_handles(ctx, self.member_converter, handles, maxcnt=None)
+        await self._ranklist(ctx, contest_id, handles, vc=True)
+
+    @commands.command(brief='Start a rated vc.', usage='contest_id duration handles')
+    @commands.has_any_role('Admin', 'Moderator')
+    async def ratedvc(self, ctx, contest_id:int, duration:int, *handles: str):
```
krofna, line 490: the vc should persist even if we restart the bot

tle/cogs/contests.py
```diff
@@ -461,10 +467,59 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+        await ctx.send(f'Starting ratedvc {contest_id} with handles: {handles}')
```
krofna, line 491: there should be a way to sync with CF vc since it relies on VIRTUAL subs. starting immediately is bad anyway

tle/cogs/contests.py
krofna, line 490: duration should probably be in minutes or fetched from CF

tle/cogs/contests.py
```diff
@@ -461,10 +467,59 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+        if show_contest_embed:
+            await ctx.send(embed=self._make_contest_embed_for_ranklist(ranklist))
         paginator.paginate(self.bot, ctx.channel, pages, wait_time=_STANDINGS_PAGINATE_WAIT_TIME)
 
+    @commands.command(brief='Show ranklist for the given vc, considering only the given handles')
+    @commands.has_any_role('Admin', 'Moderator')
+    async def vc_ranklist(self, ctx, contest_id: int, *handles: str):
+        handles = set(handles)
+        if not handles:
+            handles.add('+server')
```
krofna, line 479: for some reason this produces `Codeforces API error`. needs to be investigated

tle/cogs/contests.py
Mohammad-Yasser: So it doesn't have to be "starting immediately". Actually, it's not aware of who started when, and multiple people can start at different times, as long as they finish before the given `duration`.
So it only generates the final ranklist which is used for updating ratings after the given duration, and during that duration if there are any submissions from the given handles, it displays a new ranklist.

tle/cogs/contests.py
Mohammad-Yasser: Yeah I already included that in the Future steps in the above comment.

tle/util/cache_system2.py
```diff
@@ -605,6 +611,22 @@ async def generate_ranklist(self, contest_id, *, fetch_changes=False, predict_ch
 
         return ranklist
 
+    async def generate_vc_ranklist(self, contest_id, handles):
+        contest, problems, standings = await cf.contest.standings(contest_id=contest_id,
```
krofna, line 615: this produces a huge request upon +server. reason for cf api error. another reason to track vcs like duels.

tle/cogs/contests.py
Mohammad-Yasser: I agree to make it in mins, maybe postpone fetching it from CF to later, till it's stable and well tested, because we may need to change the duration to like 10 mins for testing, or more than the contest's duration to give some buffer just in case, it won't harm anyway.

tle/cogs/contests.py
krofna, line 491: also output should be nicer here. this is rough. make an embed maybe.

tle/util/cache_system2.py
Mohammad-Yasser: Hmm maybe if the given handles are too many, then just fetch all the standings and then filter the response by the given handles, what do you think?

Anyway, it's not intended at all to use `+server` with this, I don't think a `ratedvc` will have that so many participants. I added `;vcranklist` just for Admins to debug sth or so, but it should not needed to be done in itself, since that `;ratedvc` already watches the vc and generates the ranklists.

krofna: sure.

tle/cogs/contests.py
krofna: oh i get it. it should be documented somehow. either in help or in message "you have xhr ym to complete a virtual contest or smthing

Mohammad-Yasser:
As per suggested by @krofna , compute the rating change per contestant as if he entered the contest live.

tle/cogs/contests.py
```diff
@@ -467,12 +494,212 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+    @staticmethod
+    def _make_vc_rating_changes_embed(guild, contest_id, change_by_handle):
+        """Make an embed containing a list of rank changes and top rating increases for the members
+        of this guild.
+        """
+        contest = cf_common.cache2.contest_cache.get_contest(contest_id)
+        user_id_handle_pairs = cf_common.user_db.get_handles_for_guild(guild.id)
+        member_handle_pairs = [(guild.get_member(int(user_id)), handle)
+                               for user_id, handle in user_id_handle_pairs]
+        def ispurg(member):
```
krofna, line 535: this check shouldnt be here. purgatory shouldnt even be allowed to create vcs because it could be abused and they don't gain anything from it anyway.

tle/cogs/contests.py
```diff
@@ -467,12 +494,212 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+
+    @commands.command(brief='Plot vc rating for a single user', usage = '@user')
+    async def vc_rating(self, ctx, member: discord.Member):
+        """Plots VC rating for a single user."""
+        rating_history = cf_common.user_db.get_vc_rating_history(member.id)
+        if not rating_history:
+            raise ContestCogError(f'Nothing to plot.')
+
+        plot_data = defaultdict(list)
+        plot_data[member.display_name].append((0, 1500))
```
krofna, line 662: this data point shouldnt exist just like on cf

tle/cogs/contests.py
```diff
@@ -467,12 +494,212 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+        rank_to_role = {role.name: role for role in guild.roles}
+
+        def rating_to_displayable_rank(rating):
+            rank = cf.rating2rank(rating).title
+            role = rank_to_role.get(rank)
+            return role.mention if role else rank
+
+        rank_changes_str = []
+        for member, change in member_change_pairs:
+            cache = cf_common.cache2.rating_changes_cache
```
krofna, line 552: i am confused. you are looking at cf rating instead of vc rating

tle/cogs/contests.py
```diff
@@ -467,12 +494,212 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+            if (change.oldRating == 1500
```
krofna, line 553: redundant check?

tle/cogs/contests.py
```diff
@@ -467,12 +494,212 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+            rating_change_str = (f'{member.mention} [{change.handle}]({cf.PROFILE_BASE_URL}{change.handle}): {change.oldRating} '
+                            f'\N{HORIZONTAL BAR} **{delta:+}** \N{LONG RIGHTWARDS ARROW} '
+                            f'{change.newRating}')
+            rating_changes_str.append(rating_change_str)
+
+        desc = '\n'.join(rank_changes_str) or 'No rank changes'
+        embed = discord_common.cf_color_embed(title=contest.name, url=contest.url, description=desc)
+        embed.set_author(name='VC Results')
+        embed.add_field(name='Rating Changes',
+                        value='\n'.join(rating_changes_str) or 'No rating changes',
```
krofna, line 580: rating_changes_str should never be empty

tle/cogs/contests.py
```diff
@@ -467,12 +494,212 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+        member_ids = cf_common.user_db.get_rated_vc_user_ids(vc_id)
+        handles = [cf_common.user_db.get_handle(member_id, channel.guild.id) for member_id in member_ids]
+        handle_to_member_id = {handle : member_id for handle, member_id in zip(handles, member_ids)}
+        now = time.time()
+        if now < vc.finish_time:
+            # Display current standings
+            ranklist = await cf_common.cache2.ranklist_cache.generate_vc_ranklist(vc.contest_id, handle_to_member_id)
+            await self._ranklist(channel, vc.contest_id, handles, vc=True, show_contest_embed=True, vc_start_time=vc.start_time, vc_end_time=vc.finish_time, ranklist=ranklist, delete_after=_WATCHING_RATED_VC_WAIT_TIME)
+            return
+        ranklist = await cf_common.cache2.ranklist_cache.generate_vc_ranklist(vc.contest_id, handle_to_member_id)
```
krofna, line 595: problem: if there are still submissions judging (e.g. queue) this will be wrong. you should wait until everything is settled.

tle/cogs/contests.py
```diff
@@ -467,12 +494,212 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+            if delta is None: # The user did not participate.
+                delta = 0 # TODO check if there's a better way to handle this case.
+            old_rating = cf_common.user_db.get_vc_rating(member_id)
+            new_rating = old_rating + delta
+            rating_change_by_handle[handle] = RatingChange(handle=handle, oldRating=old_rating, newRating=new_rating)
+            cf_common.user_db.update_vc_rating(vc_id, member_id, new_rating)
+        cf_common.user_db.finish_rated_vc(vc_id)
+        await channel.send(embed=self._make_vc_rating_changes_embed(channel.guild, vc.contest_id, rating_change_by_handle))
+        await self._ranklist(channel, vc.contest_id, handles, vc=True, ranklist=ranklist, show_contest_embed=False)
+        return
```
krofna, line 609: redundant

tle/cogs/contests.py
```diff
@@ -467,12 +494,212 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+        paginator.paginate(self.bot, ctx.channel, pages, wait_time=5 * 60, set_pagenum_footers=True)
+
+    @commands.command(brief='Plot vc rating for a single user', usage = '@user')
+    async def vc_rating(self, ctx, member: discord.Member):
+        """Plots VC rating for a single user."""
+        rating_history = cf_common.user_db.get_vc_rating_history(member.id)
+        if not rating_history:
+            raise ContestCogError(f'Nothing to plot.')
+
+        plot_data = defaultdict(list)
```
krofna, line 661: i dont get it. you create a dict but you only allow a single user. either allow multiple users (preferred) or just plot rating_history without the extra step

tle/cogs/contests.py
```diff
@@ -467,12 +494,212 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+
+    async def _watch_rated_vc(self, vc_id: int, channel):
+        vc = cf_common.user_db.get_rated_vc(vc_id)
+        member_ids = cf_common.user_db.get_rated_vc_user_ids(vc_id)
+        handles = [cf_common.user_db.get_handle(member_id, channel.guild.id) for member_id in member_ids]
+        handle_to_member_id = {handle : member_id for handle, member_id in zip(handles, member_ids)}
+        now = time.time()
+        if now < vc.finish_time:
+            # Display current standings
+            ranklist = await cf_common.cache2.ranklist_cache.generate_vc_ranklist(vc.contest_id, handle_to_member_id)
```
krofna, line 592: you do this line in both branches so just factor it out

tle/util/cache_system2.py
```diff
@@ -605,6 +604,36 @@ async def generate_ranklist(self, contest_id, *, fetch_changes=False, predict_ch
...
+                   and row.party.participantType == 'VIRTUAL']
+        current_vc_rating =  {handle: cf_common.user_db.get_vc_rating(handle_to_member_id.get(handle))
+                                for handle in handles}
+        ranklist = Ranklist(contest, problems, standings, now, is_rated=True)
+        delta_by_handle = {}
+        for handle in handles:
+            mixed_ratings = current_official_rating.copy()
+            mixed_ratings[handle] = current_vc_rating.get(handle)
+            ranklist.predict(mixed_ratings)
+            delta_by_handle[handle] = ranklist.delta_by_handle.get(handle, 0)
```
krofna, line 632: why `,0`?

tle/cogs/contests.py
```diff
@@ -467,12 +494,212 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+            await self._ranklist(channel, vc.contest_id, handles, vc=True, show_contest_embed=True, vc_start_time=vc.start_time, vc_end_time=vc.finish_time, ranklist=ranklist, delete_after=_WATCHING_RATED_VC_WAIT_TIME)
+            return
+        ranklist = await cf_common.cache2.ranklist_cache.generate_vc_ranklist(vc.contest_id, handle_to_member_id)
+        rating_change_by_handle = {}
+        RatingChange = namedtuple('RatingChange', 'handle oldRating newRating')
+        for handle, member_id in zip(handles, member_ids):
+            delta = ranklist.delta_by_handle.get(handle)
+            if delta is None: # The user did not participate.
+                delta = 0 # TODO check if there's a better way to handle this case.
```
krofna, line 601: continue?

tle/util/cache_system2.py
```diff
@@ -605,6 +604,36 @@ async def generate_ranklist(self, contest_id, *, fetch_changes=False, predict_ch
...
+        # Exclude PRACTICE and MANAGER
+        standings = [row for row in standings
+                     if row.party.participantType in ('CONTESTANT', 'OUT_OF_COMPETITION', 'VIRTUAL')]
+        _, _, official_standings = await cf.contest.standings(contest_id=contest_id)
+        now = time.time()
+        rating_changes = await cf.contest.ratingChanges(contest_id=contest_id)
+        current_official_rating = {rating_change.handle : rating_change.oldRating
+                                    for rating_change in rating_changes}
+
+        # TODO: assert that none of the given handles are in the official standings.
```
krofna, line 620: how can this happen?

Mohammad-Yasser: It should not happen if this function is triggered by `;ratedvc`, but it can be triggered by `;vc_ranklist`.

tle/util/cache_system2.py
Mohammad-Yasser: In case the user didn't make any submissions as a virtual contestant, the handle won't exist in the ranklist.

tle/cogs/contests.py
```diff
@@ -467,12 +494,212 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+        handles = await cf_common.resolve_handles(ctx, self.member_converter, handles, maxcnt=50)
+        await self._ranklist(ctx.channel, contest_id, handles, vc=True, show_contest_embed=False)
+
+    @commands.command(brief='Start a rated vc.', usage='<contest_id> <duration_in_mins> <@user1 @user2 ...>')
+    async def ratedvc(self, ctx, contest_id:int, duration:int, *members: discord.Member):
+        if not members:
+            raise ContestCogError('Missing members')
+        if duration <= 0:
+            raise ContestCogError('Duration must be positive.')
+        await cf.contest.ratingChanges(contest_id=contest_id)
```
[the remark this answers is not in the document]
Mohammad-Yasser: I did it as a check it's a rated contest, and if it's not, it raises an error.
Changed it to raise a meaningful error instead.

tle/cogs/contests.py
```diff
@@ -410,34 +414,49 @@ def _make_contest_embed_for_ranklist(ranklist):
...
-            handles.remove('+server')
-            guild_handles = [handle for discord_id, handle
-                             in cf_common.user_db.get_handles_for_guild(ctx.guild.id)]
-            handles.update(guild_handles)
-        handles = await cf_common.resolve_handles(ctx, self.member_converter, handles, maxcnt=None)
-
+        wait_msg = await channel.send('Generating ranklist, please wait...')
+        if ranklist is None:
+            if vc:
+                assert False, 'Ranklists must be pre-generated for VCs'
```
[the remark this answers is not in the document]
Mohammad-Yasser: Done, I raised a `ContestCogError`, but not sure if this is what should be raised.

tle/cogs/contests.py
```diff
@@ -410,34 +414,49 @@ def _make_contest_embed_for_ranklist(ranklist):
...
+            else:                                                                    
+                try:
+                    ranklist = cf_common.cache2.ranklist_cache.get_ranklist(contest)
+                except cache_system2.RanklistNotMonitored:
+                    if contest.phase == 'BEFORE':
+                        raise ContestCogError(f'Contest `{contest.id} | {contest.name}` has not started')
+                    ranklist = await cf_common.cache2.ranklist_cache.generate_ranklist(contest.id,
+                                                                                    fetch_changes=True)
```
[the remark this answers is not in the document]
Mohammad-Yasser: Hmm sorry but I don't get specifically which part do you mean, can you please clarify it or suggest changes?

tle/cogs/contests.py
```diff
@@ -467,12 +494,212 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+    async def ratedvc(self, ctx, contest_id:int, duration:int, *members: discord.Member):
+        if not members:
+            raise ContestCogError('Missing members')
+        if duration <= 0:
+            raise ContestCogError('Duration must be positive.')
+        await cf.contest.ratingChanges(contest_id=contest_id)
+        handles = cf_common.members_to_handles(members, ctx.guild.id)
+        
+        _, _, ranklist = await cf.contest.standings(contest_id=contest_id, handles=handles, show_unofficial=True)
+        if ranklist:
```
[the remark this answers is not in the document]
Mohammad-Yasser: I fixed it for div2 thing, but I don't get wdym by kotlin, there are unrated anyway right?

tle/cogs/contests.py
Mohammad-Yasser: So there are two ways to handle the case of the user submitting nothing during the vc:
1- Assign rating change equal to 0. This is what's implemented currently and it's pretty simple.
2- Consider the user wasn't registered for this vc. This may be better from the user's perspective because it'll mimic what CF does, but its implementation will be more complicated.

For me, I prefer 1 because I don't think the extra implementation is worthy and I don't think the user will be affected by it, like currently, the only thing that uses the vc history is the plotting, and the plotting output will be the same regardless of which option is implemented.

tle/cogs/contests.py
Mohammad-Yasser: I was planning to allow multiple users as a future step.
Implemented that now.

tle/util/db/user_db_conn.py
```diff
@@ -179,6 +223,18 @@ def _insert_many(self, table: str, columns, values: list):
...
+        res = self.conn.execute(query).fetchone()
+        self.conn.row_factory = None
+        return res
+
+    def _fetchall(self, query:str, row_factory=None):
+        self.conn.row_factory = row_factory
+        res = self.conn.execute(query).fetchall()
+        self.conn.row_factory = None
+        return res
+
```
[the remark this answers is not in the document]
Mohammad-Yasser: Because different queries need different row factories.
I use it in some queries to return the result as a namedtuple, but that won't work in other queries.

tle/util/codeforces_common.py
```diff
@@ -195,6 +218,14 @@ def days_ago(t):
...
     """Convert an iterable of strings to CF handles. A string beginning with ! indicates Discord username,
      otherwise it is a raw CF handle to be left unchanged."""
+    handles = set(handles)
+    if not handles:
+        handles.add('+server')
+    if '+server' in handles:
+        handles.remove('+server')
+        guild_handles = {handle for discord_id, handle
+                            in user_db.get_handles_for_guild(ctx.guild.id)}
+        handles.update(guild_handles)
```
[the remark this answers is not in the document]
Mohammad-Yasser: Hmm I found that this was done in every command that calls `resolve_handles` but I agree that it maybe the case where this is not desired. What about adding a `default_to_all_server` param that defaults to True?

tle/cogs/contests.py
```diff
@@ -467,12 +494,220 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+
+        handles = cf_common.members_to_handles(members, ctx.guild.id)
+        visited_contests = await cf_common.get_visited_contests(handles)
+        if contest_id in visited_contests:
+            raise ContestCogError(f'Some of the handles: {", ".join(handles)} have submissions in the contest')
+        start_time = time.time()
+        finish_time = start_time + duration * 60
+        cf_common.user_db.create_rated_vc(contest_id, start_time, finish_time, [member.id for member in members])
+        title = f'Starting rated VC {contest_id} with handles:'
+        msg = discord.utils.escape_markdown("\n".join(f'[{handle}]({cf.PROFILE_BASE_URL}{handle})' for handle in handles))
```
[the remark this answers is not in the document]
Mohammad-Yasser: I think it doesn't make a difference, as it doesn't affect URLs.

tle/cogs/contests.py
```diff
@@ -467,12 +494,220 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+        discord_file = gc.get_current_figure_as_file()
+        embed = discord_common.cf_color_embed(title='VC rating graph')
+        discord_common.attach_image(embed, discord_file)
+        discord_common.set_author_footer(embed, ctx.author)
+        await ctx.send(embed=embed, file=discord_file)
 
     @discord_common.send_error_if(ContestCogError, rl.RanklistError,
-                                  cache_system2.CacheError,  cf_common.ResolveHandleError)
+                                  cache_system2.CacheError,  cf_common.ResolveHandleError,
+                                  commands.errors.MissingRequiredArgument)
```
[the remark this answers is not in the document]
Mohammad-Yasser: Hmm so do you agree that this is needed or you wanna discuss it first with @bot-dev ?
I don't see why it may be undesired.

tle/cogs/contests.py
```diff
@@ -467,12 +494,220 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+        await channel.send(embed=self._make_vc_rating_changes_embed(channel.guild, vc.contest_id, rating_change_by_handle))
+        await self._ranklist(channel, vc.contest_id, handles, vc=True, ranklist=ranklist, show_contest_embed=False)
+
+    @tasks.task_spec(name='WatchRatedVCs',
+                     waiter=tasks.Waiter.fixed_delay(_WATCHING_RATED_VC_WAIT_TIME))
+    async def _watch_rated_vcs_task(self, _):
+        ongoing_rated_vcs = cf_common.user_db.get_ongoing_rated_vc_ids()
+        if ongoing_rated_vcs is None:
+            return
+        channel_id = os.environ.get('RATED_VC_CHANNEL_ID')
```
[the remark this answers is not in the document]
krofna: @Mohammad-Yasser look how it's done in contests.py (`;remind here`)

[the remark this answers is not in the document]
Mohammad-Yasser: So after doing this, I realized that that it will break `ratedvc`s because they are not guild-aware, so where should `ratedvc`s be posted? It'll be needed first to make the `ratedvc`s guild-aware.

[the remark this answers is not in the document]
Mohammad-Yasser: Ok now `ratedvc` is guild-aware.

tle/util/codeforces_common.py
```diff
@@ -195,6 +218,14 @@ def days_ago(t):
 async def resolve_handles(ctx, converter, handles, *, mincnt=1, maxcnt=5):
     """Convert an iterable of strings to CF handles. A string beginning with ! indicates Discord username,
      otherwise it is a raw CF handle to be left unchanged."""
+    handles = set(handles)
+    if not handles:
+        handles.add('+server')
```
krofna, line 223: this part should most certainly not be part of every command

Mohammad-Yasser: Added a parameter `default_to_all_server` which is `False` by default. If it is `True` and `handles` is empty, it uses `+server`. Any problem with this?

tle/cogs/contests.py
```diff
@@ -461,18 +482,243 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
 
         problem_indices = [problem.index for problem in ranklist.problems]
         pages = self._make_standings_pages(contest, problem_indices, handle_standings, deltas)
+        paginator.paginate(self.bot, channel, pages, wait_time=_STANDINGS_PAGINATE_WAIT_TIME, delete_after=delete_after)
+        
+
+    @commands.command(brief='Show ranklist for the given vc, considering only the given handles', usage = '<contest_id> [handles]')
+    async def vc_ranklist(self, ctx, contest_id: int, *members: discord.Member):
+        member_ids = [member.id for member in members]
+        handles = [cf_common.user_db.get_handle(member_id, channel.guild.id) for member_id in member_ids]
```
krofna, line 491: channel is undefined

tle/cogs/contests.py
```diff
@@ -461,18 +482,243 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+        for member, change in member_change_pairs:
+            if len(cf_common.user_db.get_vc_rating_history(member.id)) == 1:
+                # If this is the user's first rated contest.
+                # TODO handle the new rating system changes regarding newcomers.
+                old_role = 'Unrated'
+            else:
+                old_role = rating_to_displayable_rank(change.oldRating)
+            new_role = rating_to_displayable_rank(change.newRating)
+            if new_role != old_role:
+                rank_change_str = (f'{member.mention} [{discord.utiles.escape_markdown(change.handle)}]({cf.PROFILE_BASE_URL}{change.handle}): {old_role} '
```
krofna, line 551: typo `utiles`

tle/cogs/contests.py
```diff
@@ -461,18 +482,243 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+            rating_history = cf_common.user_db.get_vc_rating_history(member.id)
+            if not rating_history:
+                raise ContestCogError(f'{member.mention} has no vc history.')
+            for vc_id, rating in rating_history:
+                plot_data[member.display_name].append((vc_id, rating))
+
+        plt.clf()
+        # plot at least from mid gray to mid purple
+        min_rating = 1350
+        max_rating = 1550
```
krofna, line 688: these should be 1100 and 1800

tle/cogs/contests.py
```diff
@@ -461,18 +482,243 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+        members = members or (ctx.author, )
+        if len(members) > 5:
+            raise ContestCogError('Cannot plot more than 5 VCers at once.')
+        plot_data = defaultdict(list)
+        for member in members:
+            rating_history = cf_common.user_db.get_vc_rating_history(member.id)
+            if not rating_history:
+                raise ContestCogError(f'{member.mention} has no vc history.')
+            for vc_id, rating in rating_history:
+                plot_data[member.display_name].append((vc_id, rating))
```
krofna, line 683: x axis should be timestamp like rating not number like duel

tle/cogs/contests.py
```diff
@@ -461,18 +482,243 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+            plt.plot(x, y,
+                     linestyle='-',
+                     marker='o',
+                     markersize=2,
+                     markerfacecolor='white',
+                     markeredgewidth=0.5)
+
+        gc.plot_rating_bg(cf.RATED_RANKS)
+        plt.xlim(0, rating_history[-1].vc_id)
+        plt.ylim(min_rating - 100, max_rating + 100)
```
krofna, line 704: should be +200

tle/cogs/contests.py
```diff
@@ -461,18 +483,255 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+                     marker='o',
+                     markersize=4,
+                     markerfacecolor='white',
+                     markeredgewidth=0.5)
+
+        gc.plot_rating_bg(cf.RATED_RANKS)
+        plt.gcf().autofmt_xdate()
+
+        if min_date == max_date:
+            plt.xlim(min_date - dt.timedelta(days=30), max_date + dt.timedelta(days=30))
```
krofna, line 713: this is inconsistent with `;plot rating` i think its better to just dump it but if there is reason to keep it -> #bot-dev

krofna:
If contestant doesn't make a submission he shouldn't get +0 rating change instead it should be as if he did not compete at all.

tle/cogs/contests.py
```diff
@@ -461,18 +483,255 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+            plt.xlim(min_date, max_date)
+        plt.gca().xaxis.set_major_formatter(mdates.DateFormatter('%Y-%m-%d'))
+        plt.ylim(min_rating - 100, max_rating + 200)
+        labels = [
+            gc.StrWrap('{} ({})'.format(
+                member_display_name,
+                rating_data[-1][1]))
+            for member_display_name, rating_data in plot_data.items()
+        ]
+        plt.legend(labels, loc='upper left')
```
krofna, line 724: prop=gc.fontprop

tle/cogs/contests.py
krofna: just delete the user from rated_vc_users

tle/cogs/contests.py
```diff
@@ -461,18 +483,255 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+        channel_id = cf_common.user_db.get_rated_vc_channel(ctx.guild.id)
+        channel = await ctx.guild.get_channel(channel_id)
+        if channel is None:
+            raise ContestCogError('There is no rated vc channel')
+        embed = discord_common.embed_success('Current rated vc channel')
+        embed.add_field(name='Channel', value=channel.mention)
+        await ctx.send(embed=embed)
+
+    @commands.command(brief='Show vc ratings', usage = '')
+    async def vc_ratings(self, ctx):
```
krofna, line 642: remove _

tle/cogs/contests.py
```diff
@@ -461,18 +483,255 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+            return 'VC Ratings', embed
+
+        if not users:
+            raise ContestCogError('There are no active VCers.')
+
+        pages = [make_page(chunk, k) for k, chunk in enumerate(paginator.chunkify(users, _PER_PAGE))]
+        paginator.paginate(self.bot, ctx.channel, pages, wait_time=5 * 60, set_pagenum_footers=True)
+
+    @commands.command(brief='Plot vc rating for a list of at most 5 users', usage = '@user1 @user2 ..')
+    async def vc_rating(self, ctx, *members: discord.Member):
```
krofna, line 673: remove _

tle/cogs/contests.py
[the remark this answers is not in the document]
krofna: we should definitely do this for other cogs (if applicable). for example purgatory seems to be unable to figure out how to use ;visualrank. but yeah, other PR.

tle/cogs/contests.py
```diff
@@ -4,28 +4,33 @@
 import logging
 import time
 import datetime as dt
-
-from collections import defaultdict
+import os
```
krofna, line 7: is this used?

tle/util/db/user_db_conn.py
```diff
@@ -707,5 +770,116 @@ def update_status(self, active_ids: list):
...
+        self.conn.commit()
+
+    def get_rated_vc_channel(self, guild_id):
+        query = ('SELECT channel_id '
+                 'FROM rated_vc_settings '
+                 'WHERE guild_id = ?')
+        channel_id = self.conn.execute(query, (guild_id,)).fetchone()
+        return int(channel_id[0]) if channel_id else None
+
+    def remove_last_ratedvc_participation(self, user_id:str):
```
krofna, line 871: this is no good because user can be in multiple vcs. you need to pass id.

tle/cogs/contests.py
```diff
@@ -461,18 +482,254 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+                     marker='o',
+                     markersize=4,
+                     markerfacecolor='white',
+                     markeredgewidth=0.5)
+
+        gc.plot_rating_bg(cf.RATED_RANKS)
+        plt.gcf().autofmt_xdate()
+
+        if min_date != max_date:
+            plt.xlim(min_date, max_date)
```
krofna, line 713: just remove this altogether then. let it work like ;plot rating

tle/util/db/user_db_conn.py
Mohammad-Yasser: I think we should ban entering multiple vcs simultaneously.
The reason I restricted this to be only for the last vc is that because we cannot delete another one because the ratings of the ones after it depend on it.

krofna: ok then do that

tle/cogs/contests.py
Mohammad-Yasser: Hmm I don't mind that but the reason I did this was to make it like CF.

krofna: #bot-dev

tle/cogs/contests.py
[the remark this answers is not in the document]
Mohammad-Yasser: Okay, I can make another PR. I think it's better to catch the base exception [UserInputError](https://discordpy.readthedocs.io/en/latest/ext/commands/api.html#discord.ext.commands.UserInputError). What do you think?

tle/cogs/contests.py
```diff
@@ -461,18 +492,261 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+        """Plots VC rating for at most 5 users."""
+        members = members or (ctx.author, )
+        if len(members) > 5:
+            raise ContestCogError('Cannot plot more than 5 VCers at once.')
+        plot_data = defaultdict(list)
+        
+        min_rating = 1100
+        max_rating = 1800
+        min_date = dt.datetime(year=3000, month=1, day=1)
+        max_date = dt.datetime(year=2000, month=1, day=1)
```
krofna, line 702: dead variables

tle/cogs/contests.py
```diff
@@ -461,18 +492,257 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+            raise ContestCogError(error)
+
+        handles = cf_common.members_to_handles(members, ctx.guild.id)
+        visited_contests = await cf_common.get_visited_contests(handles)
+        if contest_id in visited_contests:
+            raise ContestCogError(f'Some of the handles: {", ".join(handles)} have submissions in the contest')
+        start_time = time.time()
+        finish_time = start_time + duration * 60
+        cf_common.user_db.create_rated_vc(contest_id, start_time, finish_time, ctx.guild.id, [member.id for member in members])
+        title = f'Starting rated VC {contest_id} with handles:'
```
krofna, line 535: This is uninformative and potentially misleading. Instead write: `Starting {contest name} for:`

tle/cogs/contests.py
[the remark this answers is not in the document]
krofna: We will discuss it with meooow in #bot-dev

tle/util/codeforces_common.py
```diff
@@ -109,6 +110,28 @@ def is_nonstandard_problem(problem):
     return (is_nonstandard_contest(cache2.contest_cache.get_contest(problem.contestId)) or
             problem.tag_matches(['*special']))
 
+
+async def get_visited_contests(handles:[str]):
```
krofna, line 114: this code should be shared with ;vc command

Mohammad-Yasser: I thought about that too but thought it should be in another PR.

krofna: nop, should be part of this one since this is just copy paste from there

tle/cogs/codeforces.py
```diff
@@ -361,19 +361,9 @@ async def vc(self, ctx, *args: str):
...
-
-                try:
-                    contest = cf_common.cache2.contest_cache.get_contest(sub.problem.contestId)
-                    problem_id = (sub.problem.name, contest.startTimeSeconds)
-                    for cid in problem_to_contests[problem_id]:
-                        recommendations.discard(cid)
-                except cache_system2.ContestNotFound:
-                    pass
+        # Discard contests in which user has non-CE submissions.
+        visited_contests = await cf_common.get_visited_contests(handles)
```
krofna, line 365: the whole idea is to remove user_submissions = await ... line

tle/cogs/contests.py
```diff
@@ -461,18 +492,257 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+                     waiter=tasks.Waiter.fixed_delay(_WATCHING_RATED_VC_WAIT_TIME))
+    async def _watch_rated_vcs_task(self, _):
+        ongoing_rated_vcs = cf_common.user_db.get_ongoing_rated_vc_ids()
+        if ongoing_rated_vcs is None:
+            return
+        for rated_vc_id in ongoing_rated_vcs:
+            await self._watch_rated_vc(rated_vc_id)
+
+    @commands.command(brief='Set the rated vc channel to the current channel', usage = '')
+    async def set_ratedvc_channel(self, ctx):
```
krofna, line 642: @commands.has_role('Admin')

tle/cogs/contests.py
```diff
@@ -461,18 +492,258 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+                     linestyle='-',
+                     marker='o',
+                     markersize=4,
+                     markerfacecolor='white',
+                     markeredgewidth=0.5)
+
+        gc.plot_rating_bg(cf.RATED_RANKS)
+        plt.gcf().autofmt_xdate()
+
+        plt.gca().xaxis.set_major_formatter(mdates.DateFormatter('%Y-%m-%d'))
```
krofna, line 728: remove this as well keep it consistent with plot rating. and the import line on top of file

tle/cogs/contests.py
```diff
@@ -461,18 +492,258 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+        except cf.RatingChangesUnavailableError:
+            error = f'`{contest.name}` was unrated or the ratings changes are not published yet.'
+            raise ContestCogError(error)
+
+        ongoing_vc_member_ids = _get_ongoing_vc_participants()
+        this_vc_member_ids = {str(member.id) for member in members}
+        intersection = this_vc_member_ids & ongoing_vc_member_ids
+        if intersection:
+            # TODO: Think of another name xD
+            bad_boys = ", ".join([ctx.guild.get_member(int(member_id)).mention for member_id in intersection])
```
krofna, line 524: busy_members

tle/cogs/contests.py
```diff
@@ -461,18 +492,258 @@ async def ranklist(self, ctx, contest_id: int, *handles: str):
...
+        def rating_to_displayable_rank(rating):
+            rank = cf.rating2rank(rating).title
+            role = rank_to_role.get(rank)
+            return role.mention if role else rank
+
+        rank_changes_str = []
+        for member, change in member_change_pairs:
+            if len(cf_common.user_db.get_vc_rating_history(member.id)) == 1:
+                # If this is the user's first rated contest.
+                # TODO handle the new rating system changes regarding newcomers.
```
krofna, line 564: don't

tle/cogs/contests.py
```diff
@@ -410,34 +425,46 @@ def _make_contest_embed_for_ranklist(ranklist):
...
     @commands.command(brief='Show ranklist for given handles and/or server members')
     async def ranklist(self, ctx, contest_id: int, *handles: str):
         """Shows ranklist for the contest with given contest id. If handles contains
         '+server', all server members are included. No handles defaults to '+server'.
         """
-
+        handles = await cf_common.resolve_handles(ctx, self.member_converter, handles, maxcnt=None, default_to_all_server=True)
         contest = cf_common.cache2.contest_cache.get_contest(contest_id)
-        wait_msg = None
+        wait_msg = await channel.send('Generating ranklist, please wait...')
```
krofna, line 450: ctx.channel

tle/cogs/contests.py
```diff
@@ -410,34 +425,46 @@ def _make_contest_embed_for_ranklist(ranklist):
...
-            handles.add('+server')
-        if '+server' in handles:
-            handles.remove('+server')
-            guild_handles = [handle for discord_id, handle
-                             in cf_common.user_db.get_handles_for_guild(ctx.guild.id)]
-            handles.update(guild_handles)
-        handles = await cf_common.resolve_handles(ctx, self.member_converter, handles, maxcnt=None)
+                                                                            fetch_changes=True)
+        await wait_msg.delete()
+        await channel.send(embed=self._make_contest_embed_for_ranklist(ranklist))
```
krofna, line 460: ctx.channel

tle/cogs/contests.py
Mohammad-Yasser: Without this line, it can appear like this:
image

tle/util/cache_system2.py
```diff
@@ -605,6 +604,36 @@ async def generate_ranklist(self, contest_id, *, fetch_changes=False, predict_ch
...
         return ranklist
 
+    async def generate_vc_ranklist(self, contest_id, handle_to_member_id):
+        handles = list(handle_to_member_id.keys())
+        contest, problems, standings = await cf.contest.standings(contest_id=contest_id,
+                                                                  show_unofficial=True)
+        # Exclude PRACTICE and MANAGER
+        standings = [row for row in standings
+                     if row.party.participantType in ('CONTESTANT', 'OUT_OF_COMPETITION', 'VIRTUAL')]
+        _, _, official_standings = await cf.contest.standings(contest_id=contest_id)
```
krofna, line 614: this is totally unused

tle/util/cache_system2.py
```diff
@@ -605,6 +604,36 @@ async def generate_ranklist(self, contest_id, *, fetch_changes=False, predict_ch
 
         return ranklist
 
+    async def generate_vc_ranklist(self, contest_id, handle_to_member_id):
+        handles = list(handle_to_member_id.keys())
+        contest, problems, standings = await cf.contest.standings(contest_id=contest_id,
+                                                                  show_unofficial=True)
+        # Exclude PRACTICE and MANAGER
+        standings = [row for row in standings
+                     if row.party.participantType in ('CONTESTANT', 'OUT_OF_COMPETITION', 'VIRTUAL')]
```
krofna, line 613: If you replace this line with `if row.party.participantType == 'CONTESTANT' or row.party.members[0].handle in handles` will that work as intended? i.e. virtual contestant along with officials

Mohammad-Yasser: Yes, it should work fine, but what is the advantage of this over checking `if row.party.participantType in ('CONTESTANT', 'VIRTUAL')`?

tle/util/cache_system2.py
```diff
@@ -605,6 +604,41 @@ async def generate_ranklist(self, contest_id, *, fetch_changes=False, predict_ch
...
+                                                                  show_unofficial=True)
+        # Exclude PRACTICE, MANAGER and OUR_OF_COMPETITION
+        standings = [row for row in standings
+                     if row.party.participantType == 'CONTESTANT' or
+                        row.party.members[0].handle in handles]
+        standings.sort(key=lambda row: row.rank)
+        for i in range(len(standings)):
+            standings[i] = standings[i]._replace(rank=i + 1)
+        print(standings[0].rank)
+        print(standings[0].party.members[0].handle)
```
krofna, line 619: :eyes:

tle/util/db/user_db_conn.py
```diff
@@ -707,5 +769,114 @@ def update_status(self, active_ids: list):
...
+                 'WHERE user_id = ? AND rating IS NOT NULL'
+                 )
+        ratings = self._fetchall(query, params=(user_id,), row_factory=namedtuple_factory)
+        return ratings
+
+    def set_rated_vc_channel(self, guild_id, channel_id):
+        query = '''
+            INSERT OR REPLACE INTO rated_vc_settings (guild_id, channel_id)
+            VALUES (?, ?)
+        '''
```
meooow25, line 859: Can we have uniform syntax? Would prefer to have ('strings') everywhere.

tle/util/db/user_db_conn.py
```diff
@@ -707,5 +769,114 @@ def update_status(self, active_ids: list):
...
+        id = None
+        with self.conn:
+            id = self.conn.execute(query, (contest_id, start_time, finish_time, RatedVC.ONGOING, guild_id)).lastrowid
+
+        for user_id in user_ids:
+            query = ('INSERT INTO rated_vc_users '
+                 '(vc_id, user_id) '
+                 'VALUES (? , ?)')
+            with self.conn:
+                self.conn.execute(query, (id, user_id))
```
meooow25, line 789: The whole thing is one transaction, right? Should be put entirely in a `with self.conn:`

tle/util/db/user_db_conn.py
```diff
@@ -707,5 +769,114 @@ def update_status(self, active_ids: list):
...
+        ratings = self._fetchall(query, params=(user_id,), row_factory=namedtuple_factory)
+        return ratings
+
+    def set_rated_vc_channel(self, guild_id, channel_id):
+        query = '''
+            INSERT OR REPLACE INTO rated_vc_settings (guild_id, channel_id)
+            VALUES (?, ?)
+        '''
+        self.conn.execute(query, (guild_id, channel_id))
+        self.conn.commit()
```
meooow25, line 861: Can use `with self.conn:`

tle/util/cache_system2.py
```diff
@@ -605,6 +604,39 @@ async def generate_ranklist(self, contest_id, *, fetch_changes=False, predict_ch
...
+        handles = list(handle_to_member_id.keys())
+        contest, problems, standings = await cf.contest.standings(contest_id=contest_id,
+                                                                  show_unofficial=True)
+        # Exclude PRACTICE, MANAGER and OUR_OF_COMPETITION
+        standings = [row for row in standings
+                     if row.party.participantType == 'CONTESTANT' or
+                        row.party.members[0].handle in handles]
+        standings.sort(key=lambda row: row.rank)
+        for i in range(len(standings)):
+            standings[i] = standings[i]._replace(rank=i + 1)
```
meooow25, line 617:
```py
standings = [row._replace(rank=i+1) for row in standings]
```

tle/util/db/user_db_conn.py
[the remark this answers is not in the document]
meooow25: Why not?

tle/cogs/contests.py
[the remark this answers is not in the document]
meooow25: Yeah a separate PR for UserInputError sounds good.

tle/util/codeforces_common.py
```diff
@@ -109,6 +110,28 @@ def is_nonstandard_problem(problem):
...
+            if sub.verdict == 'COMPILATION_ERROR':
+                continue
+            try:
+                contest = cache2.contest_cache.get_contest(sub.problem.contestId)
+                problem_id = (sub.problem.name, contest.startTimeSeconds)
+                for cid in problem_to_contests[problem_id]:
+                    contest_ids.add(cid)
+            except cache_system2.ContestNotFound:
+                pass
+    return contest_ids
```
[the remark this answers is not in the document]
Mohammad-Yasser: Neat :ok_hand:

tle/util/cache_system2.py
Mohammad-Yasser: You missed defining `i`?
I think it will be 
```python
standings = [row._replace(rank=i+1) for i, row in enumerate(standings)]
```
right?

krofna: yes thats what he meant

tle/util/db/user_db_conn.py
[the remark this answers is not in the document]
Mohammad-Yasser: There can be some query like `SELECT MAX(x) ...`. The namedtuple row factory will fail to make a namedtuple with field `MAX(x)` because it contains invalid characters and will result in an RTE.

PR #312, opened 2020-05-09
Adding instructions for installing pillow

Mohammad-Yasser:
Also linting readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ You will need to setup a bot on your server before continuing, follow the direct
 To start TLE export the token as an environment variable and run TLE using:
 
 ```bash
-export BOT_TOKEN="<BOT_TOKEN_FROM_DISCORD_CONSOLE>"
+export BOT_TOKEN="XXXXXXXXXXXXXXXXXXXXXXXX.XXXXXX.XXXXXXXXXXXXXXXXXXXXXXXXXXX"
 poetry run python -m tle
 ```
 
--- a/README.md
+++ b/README.md
@@ -30,7 +30,9 @@ Clone the repository
 git clone https://github.com/cheran-senthil/TLE
 ```
 
-Now all dependencies need to be installed. TLE uses [Poetry](https://poetry.eustace.io/) to manage dependencies. After installing Poetry navigate to the root of the repo and run
+### Dependencies
+
+Now all dependencies need to be installed. TLE uses [Poetry](https://poetry.eustace.io/) to manage its python dependencies. After installing Poetry navigate to the root of the repo and run
 
 ```bash
 poetry install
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,27 +1,35 @@
...
 
 ## Installation
+
 Clone the repository
+
 ```bash
 git clone https://github.com/cheran-senthil/TLE
 ```
+
 Now all dependencies need to be installed. TLE uses [Poetry](https://poetry.eustace.io/) to manage dependencies. After installing Poetry navigate to the root of the repo and run
```
algmyr, line 33:
```suggestion
### Dependencies

Now all dependencies need to be installed. TLE uses [Poetry](https://poetry.eustace.io/) to manage its python dependencies. After installing Poetry navigate to the root of the repo and run
```

README.md
```diff
@@ -39,51 +48,65 @@ python3.7 -m poetry install
 
 ---
 
-TLE also depends on cairo and pango for graphics and text rendering, which you need to install through your package manager. For Ubuntu, the relevant packages can be installed with
+TLE also depends on cairo and pango for graphics and text rendering, which you need to install through your package manager. For Ubuntu, the relevant packages can be installed with:
```
algmyr, line 51:
```suggestion
#### Library dependencies

TLE also depends on cairo and pango for graphics and text rendering, which you need to install. For Ubuntu, the relevant packages can be installed with:
```

README.md
```diff
@@ -39,51 +48,65 @@ python3.7 -m poetry install
...
+
 ```bash
 apt-get install libcairo2-dev libgirepository1.0-dev libpango1.0-dev pkg-config python3-dev gir1.2-pango-1.0
 ```
 
+TLE also depends on pillow, which needs to install the following packages:
```
algmyr, line 57:
```suggestion
Additionally TLE uses pillow for graphics, which requires the following packages:
```

README.md
```diff
@@ -39,51 +48,65 @@ python3.7 -m poetry install
...
+
+```bash
+apt-get install libjpeg-dev zlib1g-dev
+```
+
 ---
```
algmyr, line 63:
```suggestion
### Final steps
```

README.md
```diff
@@ -39,51 +48,65 @@ python3.7 -m poetry install
...
 ```bash
 export BOT_TOKEN="<BOT_TOKEN_FROM_DISCORD_CONSOLE>"
 ```
+
 and run
-```
+
+```bash
 poetry run python -m tle
 ```
```
algmyr, line 77: Seems like github doesn't like suggestions that covers deleted lines. But we could make this simper. Additionally adding something about `run.sh` might make sense. Could leave that for later though.

---

    Export the token you find in your discord console as an environment variable and run TLE using
    
    ```bash
    export BOT_TOKEN="XXXXXXXXXXXXXXXXXXXXXXXX.XXXXXX.XXXXXXXXXXXXXXXXXXXXXXXXXXX"
    poetry run python -m tle
    ```

Mohammad-Yasser: I prefer leaving `run.sh` for later because I also want to replace `environment with `.env` as per the convention or at least adding `environment` to `.gitignore`.

algmyr: Leaving it for later sounds good. I think .env is a python venv convention, so we don't want to conflict with that.

algmyr: So, the reason I wanted to rephrase this part is to make the expected format of the token very obvious. I've had to help someone debug who happened to copy the wrong token.

README.md
```diff
@@ -31,59 +41,71 @@ poetry install
...
+
+### Final steps
 
 You will need to setup a bot on your server before continuing, follow the directions [here](https://github.com/reactiflux/discord-irc/wiki/Creating-a-discord-bot-&-getting-a-token). Following this, you should have your bot appearing in your server and you should have the Discord bot token.
 
-To start TLE export the token as an environment variable
+To start TLE export the token as an environment variable and run TLE using:
+
 ```bash
 export BOT_TOKEN="<BOT_TOKEN_FROM_DISCORD_CONSOLE>"
```
Mohammad-Yasser, line 74:
```suggestion
export BOT_TOKEN="XXXXXXXXXXXXXXXXXXXXXXXX.XXXXXX.XXXXXXXXXXXXXXXXXXXXXXXXXXX"
```

PR #334, opened 2020-06-29
Fix rate issue in orzing.

algmyr:
Should be the final one (I hope).

### DIFF
--- a/tle/util/discord_common.py
+++ b/tle/util/discord_common.py
@@ -8,6 +8,7 @@ from discord.ext import commands
 
 from tle.util import codeforces_api as cf
 from tle.util import db
+from tle.util import tasks
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


tle/util/discord_common.py
```diff
@@ -74,17 +75,23 @@ async def bot_error_handler(ctx, exception):
         exc_info = type(exception), exception, exception.__traceback__
         logger.exception('Ignoring exception in command {}:'.format(ctx.command), exc_info=exc_info)
 
+class OrzPresence:
```
meooow25, line 78: A class seems overkill, you can use `task` instead of `task_spec`.

algmyr: I agree it would be cleaner. I was working on changing it, but the plumbing necessary to pass arguments to the task turned out to be annoying to make. I'll leave it as is, easy to init with the variables it needs. If you feel like doing that plumbing you're more than welcome to do it (and clean this up) in a follow-up commit. :P

meooow25: Used a closure to avoid the plumbing work.

meooow25:
Appears to be working, hopefully this fixes the issue.

PR #340, opened 2020-07-21
Add visualrank plot (#256)

### LINKED ISSUE #256: Add rank-delta plot
ParsaAlizadeh suggested adding this to the bot: https://codeforces.com/blog/entry/72719
we briefly discussed it and it might be made useful if we make it like ;plot centile; to either label certain dots (like mblazev meooow) or just somehow mark all dots which belong to server members. its like a visual ranklist

sample: 
image

ParsaAlizadeh:
Closes #256
I added command `;plot visualrank [handles...]` as discussed in #256.
It can specify handles in the plot like `;plot visualrank <contest_id> tourist`. Also, it can be zoomed using `+zoom`.
Using `+server` will just add server members to the plot (not all of codeforces users).
These are samples generated by the code:
`;plot visualrank 1368 antontrygubO_o`
plot
`;plot visualrank 1368 antontrygubO_o tourist +zoom`
zoom

### DIFF
--- a/tle/cogs/graphs.py
+++ b/tle/cogs/graphs.py
@@ -738,6 +738,94 @@ class Graphs(commands.Cog):
         discord_common.set_author_footer(embed, ctx.author)
         await ctx.send(embed=embed, file=discord_file)
 
+    @plot.command(brief='Show rating changes by rank', usage='contest_id [+server] [+zoom] [handles..]')
+    async def visualrank(self, ctx, contest_id: int, *args: str):
+        """Plot rating changes by rank. Add handles to specify a handle in the plot.
+        if arguments contains `+server`, it will include just server members and not all codeforces users.
+        Specify `+zoom` to zoom to the neighborhood of handles."""
+
+        args = set(args)
+        (in_server, zoom), handles = cf_common.filter_flags(args, ['+server', '+zoom'])
+        handles = await cf_common.resolve_handles(ctx, self.converter, handles, mincnt=0, maxcnt=20)
+
+        users = cf_common.cache2.rating_changes_cache.get_rating_changes_for_contest(contest_id)
+
+        if not users:
+            raise GraphCogError(f'No rating change cache for contest `{contest_id}`')
+
+        if in_server:
+            guild_handles = [handle for discord_id, handle
+                             in cf_common.user_db.get_handles_for_guild(ctx.guild.id)]
+            users = [user for user in users if user.handle in guild_handles]
+
+        ranks = []
+        delta = []
+        color = []
+        users_to_mark = dict()
+
+        for user in users:
+            user_delta = user.newRating - user.oldRating
+
+            ranks.append(user.rank)
+            delta.append(user_delta)
+            color.append(cf.rating2rank(user.oldRating).color_graph)
+
+            if user.handle in handles:
+                users_to_mark[user.handle] = (user.rank, user_delta)
+
+        title = users[0].contestName
+
+        plt.clf()
+        fig = plt.figure(figsize=(12, 8))
+        plt.title(title)
+        plt.xlabel('Rank')
+        plt.ylabel('Rating Changes')
+
+        ymargin = 50
+        xmargin = 50
+        if users_to_mark and zoom:
+            xmin = min(point[0] for point in users_to_mark.values())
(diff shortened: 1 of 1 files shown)


tle/cogs/graphs.py
```diff
@@ -738,6 +738,94 @@ async def country(self, ctx, *countries):
...
+    async def visualrank(self, ctx, contest_id: int, *args: str):
+        """Plot rating changes by rank. Add handles to specify a handle in the plot.
+        if arguments contains `+server`, it will include just server members and not all codeforces users.
+        Specify `+zoom` to zoom to the neighborhood of handles."""
+
+        args = set(args)
+        (in_server, zoom), handles = cf_common.filter_flags(args, ['+server', '+zoom'])
+        handles = await cf_common.resolve_handles(ctx, self.converter, handles, mincnt=0, maxcnt=20)
+
+        users = cf_common.cache2.rating_changes_cache.get_rating_changes_for_contest(contest_id)
```
meooow25, line 751: Rename `users` -> `rating_changes`.
Also CF has a habit of rolling back rating changes (just happened today as well) and the `rating_changes_cache` does not auto-fix itself, so it would be preferable to query the API instead.

tle/cogs/graphs.py
```diff
@@ -738,6 +738,94 @@ async def country(self, ctx, *countries):
...
+
+        if in_server:
+            guild_handles = [handle for discord_id, handle
+                             in cf_common.user_db.get_handles_for_guild(ctx.guild.id)]
+            users = [user for user in users if user.handle in guild_handles]
+
+        ranks = []
+        delta = []
+        color = []
+        users_to_mark = dict()
```
meooow25, line 764: Minor: Use `{}` instead of `dict()`

tle/cogs/graphs.py
```diff
@@ -738,6 +738,94 @@ async def country(self, ctx, *countries):
...
+        else:
+            ylim = 0
+            if users_to_mark:
+                ylim = max(abs(point[1]) for point in users_to_mark.values())
+            ylim = max(ylim, 200)
+            xmax = max(user.rank for user in users)
+            mark_size = 2e4 / (xmax + 2 * xmargin)
+
+            plt.xlim(-xmargin, xmax + xmargin)
+            plt.ylim(-ylim - ymargin, ylim + ymargin)
```
algmyr, line 804: You could consider setting xmin,xmax,ymin,ymax in both cases and extract the common code (L791-L794 at the time of writing). I think in the second case `xmin = 0, ymin = -ylim, ymax = ylim`.

ParsaAlizadeh:
I made the changes. I think it is ready now.

tle/cogs/graphs.py
```diff
@@ -799,7 +782,27 @@ async def visualrank(self, ctx, contest_id: int, *args: str):
...
+
+        title = rating_changes[0].contestName
+
+        plt.clf()
+        fig = plt.figure(figsize=(12, 8))
+        plt.title(title)
+        plt.xlabel('Rank')
+        plt.ylabel('Rating Changes')
+
+        mark_size = 2e4 / len(ranks)
```
ParsaAlizadeh, line 805: In the last commit, I set the marker size based on the number of visible dots in the plot. I think it gives a better visualization when using `+server` or `+zoom`. (in fact, the number of dots may much lower than the highest rank)

tle/cogs/graphs.py
```diff
@@ -738,6 +738,96 @@ async def country(self, ctx, *countries):
...
+        args = set(args)
+        (in_server, zoom), handles = cf_common.filter_flags(args, ['+server', '+zoom'])
+        handles = await cf_common.resolve_handles(ctx, self.converter, handles, mincnt=0, maxcnt=20)
+
+        rating_changes = await cf.contest.ratingChanges(contest_id=contest_id)
+        if in_server:
+            guild_handles = set(handle for discord_id, handle
+                                in cf_common.user_db.get_handles_for_guild(ctx.guild.id))
+            rating_changes = [rating_change for rating_change in rating_changes
+                              if rating_change.handle in guild_handles]
```
krofna, line 756: One last thing: if a handle which isn't in server is specified along with +server, it should also be displayed. So here `or rating_change.handle in handles`

PR #341, opened 2020-07-21
Get rid of gimme bounds

krofna:


tle/cogs/codeforces.py
```diff
@@ -104,36 +104,30 @@ async def upsolve(self, ctx, choice: int = -1):
             await ctx.send(embed=embed)
 
     @commands.command(brief='Recommend a problem',
-                      usage='[tags...] [lower] [upper]')
+                      usage='[tags...] [rating]')
     @cf_common.user_guard(group='gitgud')
     async def gimme(self, ctx, *args):
+        handle, = await cf_common.resolve_handles(ctx, self.converter, ('!' + str(ctx.author),))
+        rating = round(cf_common.user_db.fetch_cf_user(handle).effective_rating, -2)
```
meooow25, line 111: This is a bit bugged because we don't directly auto-update the user db.
For our server it's fine because we auto update roles after every contest and that updates the db.

PR #342, opened 2020-07-23
[ADD] remote branch to original repo

Abdulrahmannaser:
Contributers should pull from the original repository & push to the forked one
-  If the cloned repo is forked & upstream branch doesn't exist -> add upstream branch to Cheran's repo to pull updates
 - Else if the remote branch already exists -> pull updates from the current branch
 - Else pull from the cloned repo

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -6,7 +6,25 @@ cd "$(dirname "$0")"
 [ -e environment ] && . ./environment
 
 while true; do
-    git pull
+    
+    # Contributers should pull from the original repo & push to the forked one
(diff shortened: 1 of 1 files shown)


Abdulrahmannaser:
I found that should be done manually because I should replace `git pull upstream` with `git pull --rebase upstream` which will 

1- Roll back your repository to before your unpushed changes
2- Pull from upstream (this won't require a merge, just a fast-forward)
3- Attempt to re-apply your changes that were rolled back in step 1

But that will force you to solve conflict if appears while the script is running
So the desired feature I guess is better to be done manually

Any comment?

PR #343, opened 2020-07-25
plot cumulative solve count over time

ShafinKhadem:
**Changes:**

1. Plot cumulative solve count over time:

sample

2. Avoid unnecessary merging of problem difficulties in ;plot solved when range of difficulties <= 1500.

After change:

After

Before change:

Before

### DIFF
--- a/tle/cogs/graphs.py
+++ b/tle/cogs/graphs.py
@@ -322,7 +322,7 @@ class Graphs(commands.Cog):
             labels = [gc.StrWrap(f'{handle}: {len(ratings)}')
                       for handle, ratings in zip(handles, all_ratings)]
 
-            step = 200
+            step = 200 if filt.rhi - filt.rlo > 3000 // len(handles) else 100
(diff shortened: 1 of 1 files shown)


tle/cogs/graphs.py
```diff
@@ -322,7 +322,7 @@ async def solved(self, ctx, *args: str):
             labels = [gc.StrWrap(f'{handle}: {len(ratings)}')
                       for handle, ratings in zip(handles, all_ratings)]
 
-            step = 200
+            step = 200 if filt.rhi - filt.rlo > 1500 else 100
```
krofna, line 325: I think 3000 / len was a good idea.

ShafinKhadem:
```suggestion
            step = 200 if filt.rhi - filt.rlo > 3000//len(handles)
```

ShafinKhadem:
@krofna I have made the change.

PR #350, opened 2020-08-01
Start FAQs

aryanc403:


s-ayush2903:
@aryanc403 How do I keep this bot running always on my server? I just installed it, I have to run the script(`run.sh`) from repo root in order to run the bot in server, as I stop script, the bot too stops working. Can you please help?

algmyr:
There are a bunch of options. You could make it a system service, or you could go more low tech and just run it in a tmux/screen session. We do the latter on our instance.

PR #354, opened 2020-08-05
Catch user input errors

Mohammad-Yasser:


krofna:
needs to be rebased

tle/util/discord_common.py
```diff
@@ -71,6 +71,8 @@ async def bot_error_handler(ctx, exception):
         await ctx.send(embed=embed_alert('Sorry, this command is temporarily disabled'))
     elif isinstance(exception, cf.CodeforcesApiError):
         await ctx.send(embed=embed_alert(exception))
+    elif isinstance(exception, commands.UserInputError):
```
meooow25, line 74: You can just update the above condition to
```py
elif isinstance(exception, (commands.UserInputError, cf.CodeforcesApiError)):
    await ctx.send(embed=embed_alert(exception))
```

PR #359, opened 2020-08-11
Plot for sender when ;plot centile is called without handles

Mukundan314:
`;plot centile` without handles now plots for the sender. Plotting without markers can now be done with `;plot centile +nomarker`


cheran-senthil (commented):
lgtm, looks fairly simple to me.

PR #363, opened 2020-08-20
Add lower bound on number of rated contestants to allow ratedvc

Mohammad-Yasser:


tle/cogs/contests.py
```diff
@@ -29,6 +29,7 @@
 _FINISHED_CONTESTS_LIMIT = 5
 _WATCHING_RATED_VC_WAIT_TIME = 5 * 60  # seconds
 _RATEDVC_EXTRA_TIME = 10 * 60  # seconds
+_MIN_RATED_CONTESTANTS_FOR_RATEDVC = 50
```
meooow25, line 32: Minor: `RATED_VC`

meooow25 (approved):

tle/cogs/contests.py
Mohammad-Yasser: It's my bad, but I didn't give attention to consistency regarding that before, and hence there are some `RATEDVC` and some `RATED_VC`.

I think it's better to be consistent, but not sure which way to go. I think we all agree that in commands, `ratedvc` should be used.

I think that we can only change the constants to use `RATED_VC` and leave other usages as they are. What do you think?

meooow25: Yeah, either is fine as long as its consistent. I suggested `RATED_VC` because I saw `_WATCHING_RATED_VC_WAIT_TIME` abovee.

PR #364, opened 2020-08-29
Implement problem tag exclusion for plot commands

Mukundan314:
tags can be excluded by adding `-{tag}` to the command for example `;plot hist +dp -brute` will include dp and exclude brute force problems. the `solved`, `hist`, `curve` and `scatter` plot commands are supported


krofna:
Please exclude the binary files from the PR...

Mukundan314:
@krofna excluded the binary files

tle/util/codeforces_common.py
```diff
@@ -311,6 +312,10 @@ def parse(self, args):
                 if len(arg) == 1:
                     raise ParamParseError('Problem tag cannot be empty.')
                 self.tags.append(arg[1:])
+            elif arg[0] == '-':
```
meooow25, line 315: "-" is not suitable because Codeforces handles can have "-" in them and they will start being considered tags.
This is why we use "+" to begin with.

Mukundan314: made it use "e+" instead of "-"

Mukundan314:
Misunderstood https://discordapp.com/channels/555883512952258563/560991726505033748/747915058956337162 and made the PR. I don't see much use for this, so closing this PR.

PR #376, opened 2020-09-14
Bugfix/support unrated users duel

lior5654:
Added support for a duel with unrated users


lior5654:
apparently devs want to use master instead of develop, weird decision,

Issue #335, opened 2020-06-30
;_updatestatus breaks due to multiple guilds


aryanc403:
![image](https://user-images.githubusercontent.com/32961797/86182466-a696d480-bb4d-11ea-8475-c8101eba3eac.png)

krofna:
rofl

Issue #57, opened 2019-06-16
automate _updatestatus


krofna:
i.e. what ;_updatestatus does

krofna:
https://discordpy.readthedocs.io/en/latest/api.html#discord.on_member_join

restore status and role to returning members

PR #391, opened 2020-10-06
fix updatestatus

krofna:
this is some old code i had sitting locally it probably works tho


krofna:
this is a stopgap. see #57
also depends on #393

krofna:
TODO: also account for #335 by only resetting those in current guild.

PR #401, opened 2020-10-24
Update pyproject.toml

algmyr:
Restrict discord dependency to < 1.5 for until we fix issues. Enforce requirement on seaborn 0.10.1 or newer.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,10 +1,10 @@
 [[package]]
-category = "main"
-description = "multi backend asyncio cache"
 name = "aiocache"
+version = "0.11.1"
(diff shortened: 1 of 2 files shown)


PotatoHashing:
Noob question and please let me know if this is better asked offline (i.e. Discord).

Should you also run `poetry install` or another appropriate command to update the poetry.lock file?

meooow25:
To update the lock file, [`poetry update`](https://python-poetry.org/docs/cli/#update).
Deleting the lock file and doing `poetry install` will also prepare a fresh lock file.

PR #407, opened 2020-11-06
Added minimum rating for gitgud, to 1000. Could maybe be 1100

AntohiRobert:
Hey, so I added a minimum rating for gitgud. It is especially useful for beginners, to be able to solve problems in the 800-1300 range. I set it to 1000 currently, 1100 would allow for 800-1400, but it could be discouraging for beginners, maybe.


algmyr:
To summarize some stuff from discord
image
image


@meooow25 @krofna Give opinions if we want to do this, and if we do how to implement it.

I think it's a very reasonable thing to want. My thoughts on implementation is in the first image.

meooow25:
Sounds reasonable to me, since we don't have real ratings.
I'm not too familiar with the gitgud stuff though, @krofna?

Issue #408, opened 2020-11-06
Nonetype object has no attribute id on ;meta guilds


dgcnz:
I just installed TLE on a Ubuntu 20.04 DigitalOcean Droplet and I'm having trouble with the `;handle` command. It registers users correctly but when trying `;handle list` I get the error "No members with registered handles." I checked poetry.lock and discord.py is correctly set on version 1.4.2 so I'm unsure about what's happening.

Also, I'm getting an error on `;meta guilds` which might be relevant:

```
06-11-2020 18:29:05:ERROR:tle.util.discord_common:Ignoring exception in command meta guilds:
Traceback (most recent call last):
  File "/root/.cache/pypoetry/virtualenvs/tle-F1GuQE82-py3.8/lib/python3.8/site-packages/discord/ext/commands/core.py", line 85, in wrapped
    ret = await coro(args, **kwargs)
  File "/root/dev/programs/TLE/tle/cogs/meta.py", line 95, in guilds
    msg = [f'Guild ID: {guild.id} | Name: {guild.name} | Owner: {guild.owner.id} | Icon: {guild.icon_url}'
  File "/root/dev/programs/TLE/tle/cogs/meta.py", line 95, in <listcomp>
    msg = [f'Guild ID: {guild.id} | Name: {guild.name} | Owner: {guild.owner.id} | Icon: {guild.icon_url}'
AttributeError: 'NoneType' object has no attribute 'id'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/root/.cache/pypoetry/virtualenvs/tle-F1GuQE82-py3.8/lib/python3.8/site-packages/discord/ext/commands/bot.py", line 903, in invoke
    await ctx.command.invoke(ctx)
  File "/root/.cache/pypoetry/virtualenvs/tle-F1GuQE82-py3.8/lib/python3.8/site-packages/discord/ext/commands/core.py", line 1302, in invoke
    await ctx.invoked_subcommand.invoke(ctx)
  File "/root/.cache/pypoetry/virtualenvs/tle-F1GuQE82-py3.8/lib/python3.8/site-packages/discord/ext/commands/core.py", line 850, in invoke
    await injected(ctx.args, **ctx.kwargs)
  File "/root/.cache/pypoetry/virtualenvs/tle-F1GuQE82-py3.8/lib/python3.8/site-packages/discord/ext/commands/core.py", line 94, in wrapped
    raise CommandInvokeError(exc) from exc
[...]

PR #409, opened 2020-11-07
Updated to Discord 1.5.0

dgcnz:
These are the changes that were referred in #408. I've tested it so far with a few commands and it seems to be working, but it is far from an exhaustive test.

### DIFF
--- a/tle/__main__.py
+++ b/tle/__main__.py
@@ -3,6 +3,7 @@ import asyncio
 import distutils.util
 import logging
 import os
+import discord
 from logging.handlers import TimedRotatingFileHandler
 from os import environ
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


tle/__main__.py
```diff
@@ -57,8 +58,9 @@ def main():
         constants.ALLOW_DUEL_SELF_REGISTER = bool(distutils.util.strtobool(allow_self_register))
 
     setup()
-
-    bot = commands.Bot(command_prefix=commands.when_mentioned_or(';'))
+    
+    intents = discord.Intents.all()
```
algmyr, line 62: Should we do `all()` or something more specific? Seems like `all()` is currently default + presences and members. We could specify that explicitly so we don't pull in extra stuff in the future. If I read the docs correctly it would be
```
intents = discord.Intents.default()
intents.presences = True
intents.members = True
```
Would be good to have opinions on other devs on this @meooow25 @krofna

algmyr (changes requested):

tle/__main__.py
meooow25: I don't think we need presence, just members would do.
Also ideally we could change the code to not depend on cached members, but that would need some digging and this is good enough to fix the issue a lot of people seem to be having, so I'm okay with it.

dgcnz: Added Intents.default() + members, seems to be working fine.

meooow25 (approved):

algmyr (approved):

algmyr:
Thanks!

PR #418, opened 2020-12-26
Paginate ;stalk. Closes #385

### LINKED ISSUE #385: paginate ;stalk
make sure to limit number of pages

krofna:
idk what should page limit be

### DIFF
--- a/tle/cogs/codeforces.py
+++ b/tle/cogs/codeforces.py
@@ -159,16 +159,21 @@ class Codeforces(commands.Cog):
         else:
             submissions.sort(key=lambda sub: sub.creationTimeSeconds, reverse=True)
 
-        msg = '\n'.join(
-            f'[{sub.problem.name}]({sub.problem.url})\N{EN SPACE}'
-            f'[{sub.problem.rating if sub.problem.rating else "?"}]\N{EN SPACE}'
-            f'({cf_common.days_ago(sub.creationTimeSeconds)})'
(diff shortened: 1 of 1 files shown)


tle/cogs/codeforces.py
```diff
@@ -159,16 +159,21 @@ async def stalk(self, ctx, *args):
...
-            f'[{sub.problem.rating if sub.problem.rating else "?"}]\N{EN SPACE}'
-            f'({cf_common.days_ago(sub.creationTimeSeconds)})'
-            for sub in submissions[:10]
-        )
-        title = '{} solved problems by `{}`'.format('Hardest' if hardest else 'Recently',
-                                                    '`, `'.join(handles))
-        embed = discord_common.cf_color_embed(title=title, description=msg)
-        await ctx.send(embed=embed)
+        def make_line(sub):
+            line = (f'[{sub.problem.name}]({sub.problem.url})\N{EN SPACE}'
```
algmyr, line 163: Nit: You can do this whole think on one line, the extra assignment provides no value. Also, there is the option of doing a `'\N{EN SPACE}'.join` which at least for me shows intention better. In that case you could even do something like
```
def make_line(...):
    parts = (a, b, c)
    return '\N{EN SPACE}'.join(parts)
```

algmyr (changes requested):

algmyr (approved):

PR #420, opened 2020-12-27
Add +nolegend. Fixes #265

### LINKED ISSUE #265: make option to not show legend in plots because it covers data sometimes
probably depends on #259

krofna:


tle/cogs/graphs.py
```diff
@@ -169,8 +169,9 @@ def scatter_outline(*args, **kwargs):
                         s=32, marker='X',
                         color=unsolvedcolor)
 
-    plt.legend(title=f'{handle}: {rating}', title_fontsize=plt.rcParams['legend.fontsize'],
-               loc='upper left').set_zorder(20)
+    if not nolegend:
```
algmyr, line 172: Nit: Having negative booleans in code is not the best thing. You could negate at parse time so that `+nolegend` sets `legend=False` or something similar.

algmyr (approved):

PR #427, opened 2021-01-05
Help format

algmyr:
Remove single newlines in help strings (that are generated from the docstrings). Newlines prefixed with `\\` remain, double newlines (paragraphs) remain. (May or may not behave akin to LaTeX now.)

### DIFF
--- a/tle/util/discord_common.py
+++ b/tle/util/discord_common.py
@@ -1,7 +1,8 @@
 import asyncio
-import logging
 import functools
+import logging
 import random
+import re
 
 import discord
 from discord.ext import commands
@@ -15,7 +16,6 @@ logger = logging.getLogger(__name__)
 _CF_COLORS = (0xFFCA1F, 0x198BCC, 0xFF2020)
 _SUCCESS_GREEN = 0x28A745
 _ALERT_AMBER = 0xFFBF00
-
 
 def embed_neutral(desc, color=discord.Embed.Empty):
     return discord.Embed(description=str(desc), color=color)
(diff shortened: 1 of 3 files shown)


tle/util/discord_common.py
```diff
@@ -129,3 +129,31 @@ async def presence_task(_):
...
+            self.paginator.add_line(signature, empty=True)
+
+        if command.help:
+            try:
+                new_help = TLEHelpCommand._reformat(command.help)
+                self.paginator.add_line(new_help, empty=True)
+            except RuntimeError:
+                for line in command.help.splitlines():
+                    self.paginator.add_line(line)
+                self.paginator.add_line()
```
meooow25, line 159: Probably a bit fragile to depend on the internal implementation, don't you think? 🤔
What if we instead have a decorator on the methods that update the docstring? Or if we don't want to be that granular, we could update `command.help` for each command in the bot after we have added the cogs.

algmyr: Is it that fragile? We could fork the thing if that makes you happier. This just felt like the minimal effort route, and I doubt the api of the DefaultHelpCommand would change drastically. What internal implementation do I actually depend on?

> What if we instead have a decorator on the methods that update the docstring?

What do you mean exactly?

> Or if we don't want to be that granular, we could update command.help for each command in the bot after we have added the cogs.

That feels hackier than what I did here tbh.

meooow25: > What internal implementation do I actually depend on?

That was a bad choice of words, I meant that a piece of logic is copied from the default impl. Also the API will probably not change so this may be fine 🤷 

> > What if we instead have a decorator on the methods that update the docstring?

> What do you mean exactly?
```py
def reformat_help(func):
    func.__doc__ = _reformat(func.__doc__)
    return func
```

> >  Or if we don't want to be that granular, we could update command.help for each command in the bot after we have added the cogs.

> That feels hackier than what I did here tbh.

Why though :waturr:
This is just the line `new_help = TLEHelpCommand._reformat(command.help)` from the PR.
```py
for command in bot.walk_commands():
    command.help = _reformat(command.help)
```

I probably like the last option the most. What do you think? 👀

PR #429, opened 2021-01-12
Fix unmagic failure for case changes

algmyr:
Just looking at the redirect you don't detect case changes.
Added logic to also check case changes and refactored logic a bit.

### DIFF
--- a/tle/cogs/handles.py
+++ b/tle/cogs/handles.py
@@ -473,17 +473,13 @@ class Handles(commands.Cog):
                     lower_map = {h.lower(): h for h in handle_chunk}
 
                     # Users could still have changed capitalization
-                    def needs_case_fix(user):
-                        return (not user.handle in handle_chunk
-                                and user.handle.lower() in lower_map)
-
-                    for user in users:
-                        if needs_case_fix(user):
+                    for handle, user in zip(handle_chunk, users):
+                        assert handle.lower() == user.handle.lower()
+                        if handle != user.handle:
                             handle = lower_map[user.handle.lower()]
                             to_fix.append((rev_lookup[handle], handle))
                     break
                 except cf.HandleNotFoundError as e:
-                    # Handle not found
                     to_fix.append((rev_lookup[e.handle], e.handle))
                     handle_chunk.remove(e.handle)
         return to_fix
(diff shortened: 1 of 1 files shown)


tle/cogs/handles.py
```diff
@@ -437,66 +437,80 @@ async def unmagic(self, ctx):
...
+        # Return summary embed
+        lines = []
+        if not fixed and not failed:
+            return discord_common.embed_success('No handles updated')
+        if fixed:
+            lines.append('**Fixed**')
+            lines += (f'{old} -> {new}' for old, new in fixed)
+        if failed:
+            lines.append('**Failed**')
+            lines += (f'{handle}' for handle in failed)
```
meooow25, line 512: `lines += failed` 👀

tle/cogs/handles.py
```diff
@@ -437,66 +437,80 @@ async def unmagic(self, ctx):
...
+        return discord_common.embed_success('\n'.join(lines))
```
meooow25, line 513: Note: If there are too many and it exceeds char limit, sending will fail 🙃

tle/cogs/handles.py
```diff
@@ -437,66 +437,80 @@ async def unmagic(self, ctx):
...
-                    await cf.user.info(handles=handles)
-                    n_same += len(handles)
-                    break  # no failed handles
+                    users = await cf.user.info(handles=handle_chunk)
+                    lower_map = {h.lower(): h for h in handle_chunk}
+
+                    # Users could still have changed capitalization
+                    def needs_case_fix(user):
+                        return (not user.handle in handle_chunk
+                                and user.handle.lower() in lower_map)
```
meooow25, line 478: > `not user.handle in handle_chunk`

Not sure I understand this.
Also `user.handle.lower()` is expected to be present in `lower_map` even if handle doesn't change, right?

meooow25 (changes requested):

tle/cogs/handles.py
algmyr: I'm aware, I really hope we won't have quite that many users switching handles :|

tle/cogs/handles.py
algmyr: yay for copy paste coding, fixed

tle/cogs/handles.py
algmyr: Discussed offline and logic updated and simplified.

PR #432, opened 2021-01-15
Implementation of issue #249

ArthurConmy:
Just realised @krofna that you self-assigned this. Hope we can work with this pull request anyway :) 

Calls to `;cfdistrib` with `[countries..]` do require API calls to CF for literally every user - is this problematic?


krofna:
For reference: https://github.com/cheran-senthil/TLE/commit/77bdbab245837febc5e575d3be17962983da3185 should be used to unmagic cache in order to make this command usable

krofna:
unmagic issue was resolved but re-downloading all the data for each invocation is slow and prone to abuse. It is necessary to cache all user profiles and not just the server members (which is done by ;roleupdate now)

PR #436, opened 2021-01-18
Improve logging output

### LINKED ISSUE #404: Add message links to loging when a command fails
I think having a message link and maybe a copy of the command issued would help with debugging why some command failed. Also sometimes I look through a debugging channel for issues and see an error but I'm not sure what command caused it.

Tangentially related, maybe allow users to turn on/off more verbose logs with a flag?

PotatoHashing:
This change

- Wraps logging output in a code block to prevent discord formatting on tracebacks.
- Adds embed for link and command that caused an exception.
- Remove logging from ;cache command because it double logged to channel where the command was sent, and in logging channel.
- Remove code block formatting from logging start up.

Fixes #404 

**Before**:
image
image
**After**:
image
image

### DIFF
--- a/tle/cogs/cache_control.py
+++ b/tle/cogs/cache_control.py
@@ -84,17 +84,7 @@ class CacheControl(commands.Cog):
         await ctx.send(f'Done, fetched {count} problems')
 
     async def cog_command_error(self, ctx, error):
-        if isinstance(error, commands.CommandInvokeError):
-            error = error.__cause__
-        lines = traceback.format_exception(type(error), error, error.__traceback__)
-        msg = '\n'.join(lines)
-        discord_msg_char_limit = 2000
-        char_limit = discord_msg_char_limit - 2 * len('```')
-        too_long = len(msg) > char_limit
-        msg = msg[:char_limit]
-        await ctx.send(f'```{msg}```')
-        if too_long:
-            await ctx.send('Check logs for full stack trace')
+        pass
 
 
 def setup(bot):
--- a/tle/cogs/cache_control.py
+++ b/tle/cogs/cache_control.py
@@ -83,9 +83,6 @@ class CacheControl(commands.Cog):
             count = await cf_common.cache2.problemset_cache.update_for_contest(contest_id)
         await ctx.send(f'Done, fetched {count} problems')
 
-    async def cog_command_error(self, ctx, error):
-        pass
-
 
 def setup(bot):
     bot.add_cog(CacheControl(bot))
--- a/tle/cogs/logging.py
+++ b/tle/cogs/logging.py
@@ -40,10 +40,10 @@ class Logging(commands.Cog, logging.Handler):
                 break
             try:
                 msg = self.format(record)
-                # Not all errors have jump urls.
+                # Not all errors will have message_contents or jump urls.
                 try:
                     await channel.send(
-                        '```Original Command: {}\nJump Url: {}```'.format(
+                        'Original Command: {}\nJump Url: {}'.format(
                             record.message_content, record.jump_url))
                 except AttributeError:
                     pass
--- a/tle/cogs/logging.py
+++ b/tle/cogs/logging.py
@@ -36,15 +36,16 @@ class Logging(commands.Cog, logging.Handler):
             if channel is None:
                 # Channel no longer exists.
                 root_logger.removeHandler(self)
-                self.logger.warning(
-                    'Logging channel not available, disabling Discord log handler.')
+                self.logger.warning('Logging channel not available, disabling Discord log handler.')
                 break
             try:
                 msg = self.format(record)
-                await channel.send('```Original Command: {}```'.format(record.message_content))
                 # Not all errors have jump urls.
                 try:
-                    channel.send('```Jump Url: {}```'.format(record.jump_url))
+                    await channel.send('```Original Command: {}```'.format(
+                        record.message_content))
+                    await channel.send('```Jump Url: {}```'.format(
+                        record.jump_url))
                 except AttributeError:
                     pass
                 discord_msg_char_limit = 2000
@@ -70,8 +71,7 @@ class Logging(commands.Cog, logging.Handler):
 def setup(bot):
     logging_cog_channel_id = os.environ.get('LOGGING_COG_CHANNEL_ID')
     if logging_cog_channel_id is None:
-        logger.info(
-            'Skipping installation of logging cog as logging channel is not provided.')
+        logger.info('Skipping installation of logging cog as logging channel is not provided.')
         return
 
     logging_cog = Logging(bot, int(logging_cog_channel_id))
--- a/tle/cogs/logging.py
+++ b/tle/cogs/logging.py
@@ -42,10 +42,9 @@ class Logging(commands.Cog, logging.Handler):
                 msg = self.format(record)
                 # Not all errors have jump urls.
                 try:
-                    await channel.send('```Original Command: {}```'.format(
-                        record.message_content))
-                    await channel.send('```Jump Url: {}```'.format(
-                        record.jump_url))
+                    await channel.send(
+                        '```Original Command: {}\nJump Url: {}```'.format(
+                            record.message_content, record.jump_url))
                 except AttributeError:
                     pass
                 discord_msg_char_limit = 2000
--- a/tle/cogs/logging.py
+++ b/tle/cogs/logging.py
@@ -36,13 +36,15 @@ class Logging(commands.Cog, logging.Handler):
             if channel is None:
                 # Channel no longer exists.
                 root_logger.removeHandler(self)
-                self.logger.warning('Logging channel not available, disabling Discord log handler.')
+                self.logger.warning(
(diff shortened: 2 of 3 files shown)


tle/cogs/cache_control.py
```diff
@@ -84,17 +84,7 @@ async def problemsets(self, ctx, contest_id):
...
-        lines = traceback.format_exception(type(error), error, error.__traceback__)
-        msg = '\n'.join(lines)
-        discord_msg_char_limit = 2000
-        char_limit = discord_msg_char_limit - 2 * len('```')
-        too_long = len(msg) > char_limit
-        msg = msg[:char_limit]
-        await ctx.send(f'```{msg}```')
-        if too_long:
-            await ctx.send('Check logs for full stack trace')
+        pass
```
meooow25, line 87: Can delete the function.

tle/cogs/logging.py
```diff
@@ -40,7 +40,18 @@ async def _log_task(self):
...
+                        embed=discord_common.embed_failed_command(record))
+                except AttributeError:
+                    pass
+                discord_msg_char_limit = 2000
+                char_limit = discord_msg_char_limit - 2 * len('```')
+                too_long = len(msg) > char_limit
+                msg = msg[:char_limit]
+                await channel.send('```{}```'.format(msg))
+                if too_long:
+                    await channel.send('`Check logs for full stack trace`')
```
meooow25, line 54: Thanks, this should have been here from the beginning.

tle/cogs/logging.py
```diff
@@ -40,7 +40,18 @@ async def _log_task(self):
                 break
             try:
                 msg = self.format(record)
-                await channel.send(msg)
+                try:
+                    await channel.send(
+                        embed=discord_common.embed_failed_command(record))
+                except AttributeError:
+                    pass
```
meooow25, line 47: What is this for? 🤔

tle/cogs/logging.py
```diff
@@ -40,7 +40,18 @@ async def _log_task(self):
                 break
             try:
                 msg = self.format(record)
-                await channel.send(msg)
+                try:
+                    await channel.send(
+                        embed=discord_common.embed_failed_command(record))
```
meooow25, line 45: I would prefer it be a simple message, embeds take up a lot of space and looks don't matter in the logging channel.

tle/cogs/cache_control.py
PotatoHashing: Done.

tle/cogs/logging.py
PotatoHashing: Acknowledged.

tle/cogs/logging.py
PotatoHashing: Added comment. Some records don't have the jump url and will throw the AttributeError.
I think it's exceptions not thrown from tle/util/discord_common.py#83

Apparently it's pythonic to ask for forgiveness instead of permission?
https://stackoverflow.com/a/610923

tle/cogs/logging.py
PotatoHashing: :blobsad: changed.

tle/cogs/logging.py
meooow25: Ah I see, was a bit hard to follow because the attr access was inside the `embed_failed_command` function.

tle/util/discord_common.py
```diff
@@ -80,12 +83,19 @@ async def bot_error_handler(ctx, exception):
...
     else:
         exc_info = type(exception), exception, exception.__traceback__
-        logger.exception('Ignoring exception in command {}:'.format(ctx.command), exc_info=exc_info)
+        logger.exception('Ignoring exception in command {}:'.format(
+            ctx.command),
+            exc_info=exc_info,
+            extra={
+            "message_content": ctx.message.content,
+            "jump_url": ctx.message.jump_url
+        })
```
meooow25, line 92: This formatting is super weird, can we rearrange things a bit? Any way you prefer so that it's nicer.

tle/cogs/logging.py
```diff
@@ -36,11 +36,24 @@ async def _log_task(self):
...
+                self.logger.warning(
+                    'Logging channel not available, disabling Discord log handler.')
                 break
             try:
                 msg = self.format(record)
-                await channel.send(msg)
+                await channel.send('```Original Command: {}```'.format(record.message_content))
+                # Not all errors have jump urls.
+                try:
+                    channel.send('```Jump Url: {}```'.format(record.jump_url))
```
meooow25, line 47: Not sure if all records will always have `message_content`. Error thrown which are caught by discord will, I think but any other place where an error is directly logged will not have it.
Also I think we could bundle this and jump url in the same message.
And missed `await` in the second send 👀

tle/cogs/logging.py
```diff
@@ -57,7 +70,8 @@ def close(self):
 def setup(bot):
     logging_cog_channel_id = os.environ.get('LOGGING_COG_CHANNEL_ID')
     if logging_cog_channel_id is None:
-        logger.info('Skipping installation of logging cog as logging channel is not provided.')
+        logger.info(
+            'Skipping installation of logging cog as logging channel is not provided.')
```
meooow25, line 74: Could you maybe revert some of these modified lines?
Adding blank lines is fine but for other stuff algmyr has a separate PR anyway so that should take care of it.

meooow25 (changes requested):

PotatoHashing:
I don't have a dev TLE anymore so I won't be able to verify these changes.

krofna:
We can make the minor changes ourselves and merge it.

tle/cogs/logging.py
PotatoHashing: Reverted the non-adding-blank-line formatting changes.

tle/cogs/logging.py
PotatoHashing: For what I've seen, errors will just have `record.message_content` = `None` which still prints in the error message just fine. Missing jump_urls caused another error to be thrown though. I think any kind of command that doesn't exist should trigger this (such as `;fake_command`) but I haven't tested extensively. 

Merged both messages into the same message inside the await.

tle/util/discord_common.py
PotatoHashing: Done.

tle/util/discord_common.py
```diff
@@ -28,16 +28,19 @@ def embed_success(desc):
...
+
 def cf_color_embed(**kwargs):
     return discord.Embed(**kwargs, color=random_cf_color())
 
+
 def set_same_cf_color(embeds):
     color = random_cf_color()
     for embed in embeds:
-        embed.color=color
+        embed.color = color
```
meooow25, line 43: Revert

tle/util/discord_common.py
```diff
@@ -116,7 +125,7 @@ async def presence(bot):
     await asyncio.sleep(60)
 
     @tasks.task(name='OrzUpdate',
-               waiter=tasks.Waiter.fixed_delay(5*60))
+                waiter=tasks.Waiter.fixed_delay(5*60))
```
meooow25, line 128: Revert

tle/cogs/logging.py
```diff
@@ -40,7 +40,20 @@ async def _log_task(self):
                 break
             try:
                 msg = self.format(record)
-                await channel.send(msg)
+                # Not all errors have jump urls.
```
meooow25, line 43: Maybe update this comment to mention message_content since the handling for that is same as jump_url?

tle/cogs/logging.py
```diff
@@ -40,7 +40,20 @@ async def _log_task(self):
...
+                try:
+                    await channel.send(
+                        '```Original Command: {}\nJump Url: {}```'.format(
+                            record.message_content, record.jump_url))
```
meooow25, line 47: okarinn reported that this makes the jump URL non-clickable.
Leaving out the backticks so its not a code block should fix that.

meooow25 (changes requested):
Everything looks good, please resolve the jump url clickable thing and we can merge.

tle/cogs/logging.py
meooow25: That's... weird. I checked and a message always has content and jump_url ([src](https://github.com/Rapptz/discord.py./blob/d01500bccd8dd87a96edb62b259856e9fa4c9ec6/discord/message.py#L338-L345)).  
Would be nice if you could look further into when that is happening, but not strictly required for this PR.

tle/util/discord_common.py
PotatoHashing: Done.

tle/util/discord_common.py
PotatoHashing: Done

tle/cogs/logging.py
PotatoHashing: Done.

tle/cogs/logging.py
PotatoHashing: Thanks okarinn for testing. Done.

meooow25 (approved):

PR #437, opened 2021-01-23
Switch cf api queries to use POST request

algmyr:
Superficial test seemed to work. Hard to check in the cases where is actually matters without painstakingly creating a huge db for testing. Testing something like `;handle unmagic_all` on server should at least verify that the limit has vastly increased.

### DIFF
--- a/tle/util/codeforces_api.py
+++ b/tle/util/codeforces_api.py
@@ -17,7 +17,6 @@ PROFILE_BASE_URL = 'https://codeforces.com/profile/'
 ACMSGURU_BASE_URL = 'https://codeforces.com/problemsets/acmsguru/'
 GYM_ID_THRESHOLD = 100000
 DEFAULT_RATING = 1500
-MAX_HANDLES_PER_QUERY = 300 # To avoid sending too large requests.
 
 logger = logging.getLogger(__name__)
 
@@ -249,7 +248,7 @@ async def _query_api(path, params=None):
         logger.info(f'Querying CF API at {url} with {params}')
         # Explicitly state encoding (though aiohttp accepts gzip by default)
         headers = {'Accept-Encoding': 'gzip'}
-        async with _session.get(url, params=params, headers=headers) as resp:
+        async with _session.post(url, params=params, headers=headers) as resp:
             try:
                 respjson = await resp.json()
             except aiohttp.ContentTypeError:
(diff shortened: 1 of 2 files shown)


tle/util/codeforces_api.py
```diff
@@ -335,14 +334,32 @@ async def problems(*, tags=None, problemset_name=None):
...
     @staticmethod
     async def info(*, handles):
-        chunks = chunkify(handles, MAX_HANDLES_PER_QUERY)
-        if len(chunks) > 1:
-            logger.warning(f'cf.info request with {len(handles)} handles,'
-            f'will be chunkified into {len(chunks)} requests.')
+        chunks = user_info_chunkify(handles)
+        #if len(chunks) > 1:
+        #    logger.warning(f'cf.info request with {len(handles)} handles,'
+        #    f'will be chunkified into {len(chunks)} requests.')
```
meooow25, line 362: Don't we want the warning any more? 🤔

tle/util/codeforces_api.py
```diff
@@ -335,14 +334,32 @@ async def problems(*, tags=None, problemset_name=None):
...
+            size = 0
+        chunk.append(handle)
+        size += len(handle) + 1
+    if chunk:
+        yield chunk
 
 class user:
     @staticmethod
     async def info(*, handles):
-        chunks = chunkify(handles, MAX_HANDLES_PER_QUERY)
```
meooow25, line 342: `chunkify` import no longer needed.

tle/util/codeforces_api.py
```diff
@@ -249,7 +248,7 @@ async def _query_api(path, params=None):
         logger.info(f'Querying CF API at {url} with {params}')
         # Explicitly state encoding (though aiohttp accepts gzip by default)
         headers = {'Accept-Encoding': 'gzip'}
-        async with _session.get(url, params=params, headers=headers) as resp:
+        async with _session.post(url, params=params, headers=headers) as resp:
```
meooow25, line 251: I think we need to change `params` to [`data`](https://docs.aiohttp.org/en/stable/client_quickstart.html#more-complicated-post-requests).

tle/util/codeforces_api.py
algmyr: I was meaning to remove is for the sake for keeping chunks as an iterator, but I guess doing list and keeping the check is fine as well.

tle/util/codeforces_api.py
algmyr: Curiously it accepts both and seem to treat them more or less equivalently. But yeah, data is the right one for post.

meooow25 (approved):

PR #443, opened 2021-02-08
Add yapf action

algmyr:


cheran-senthil:
I think we should just use the ruff we have to verify with an action instead.

algmyr:
Agreed, though this issue might predate that being added (and ruff existing) by a few years 🙃

PR #446, opened 2021-02-19
Enhance Dockerfile to make container persistent and UPD Docker.md accordingly

Shahraaz:


Docker.md
```diff
@@ -21,13 +39,23 @@ export ALLOW_DUEL_SELF_REGISTER="false"
...
+### Debugging/Running Commands inside the container
+
+To Run Commands inside the container
+
+- Get the id of the container you just run using `sudo docker ps` and run:
+
 ```console
-foo@bar:~$ sudo docker run image_id
-```
+foo@bar:~$ sudo docker exec -it container_id bash
```
algmyr, line 60: I think it's customary to mark the parts that you are supposed to replace, e.g. writing `<container_id>`

Dockerfile
```diff
@@ -1,23 +1,17 @@
 FROM ubuntu:18.04
 USER root
+WORKDIR /TLE
+
 RUN apt-get update
-RUN apt-get -y install git
+RUN apt-get install -y git apt-utils
+RUN DEBIAN_FRONTEND="noninteractive" apt-get install -y libcairo2-dev libgirepository1.0-dev libpango1.0-dev pkg-config python3-dev gir1.2-pango-1.0 python3.8-venv libpython3.8-dev libjpeg-dev zlib1g-dev python3-pip
+RUN python3.8 -m pip install poetry
```
algmyr, line 9: Should the author script (`extra/scrape_cf_contest_writers.py`) also be ran here? It's a script that really should be run occasionally just to keep the author list somewhat fresh, so not sure where the best place for it would be.

algmyr (changes requested):
Also, general comment: I presume that the DBs will now be inside the actual image? Would it be useful to give some examples as to how to access/extract them if needed?

Docker.md
Shahraaz: ok, fixed

Dockerfile
Shahraaz: It can be run, but it is better to run the script outside the container (as a normal user script). 

As long as the bot is mounted(i.e. run using `sudo docker run -v ${PWD}:/TLE -it --net host <image_id>`), the change will be reflected inside the bot as usual.

Shahraaz:
> Also, general comment: I presume that the DBs will now be inside the actual image? Would it be useful to give some examples as to how to access/extract them if needed?

As long as the bot is mounted(i.e. run using sudo docker run -v ${PWD}:/TLE -it --net host <image_id>), the change will be reflected inside the bot as usual. files will be available both inside the container and on your local machine. 

But if the user chooses to not mount the bot (not use -v ${PWD}:/TLE flag), I have already added instructions in [Docker.md](https://github.com/Shahraaz/TLE/blob/master/Docker.md#debuggingrunning-commands-inside-the-container)

Screenshot from 2021-02-20 01-05-50

Dockerfile
Shahraaz: I tried running the command while building and it executed properly too. But the file is not written anywhere, it is neither present inside the container nor outside. Better to run the script separately.

algmyr (approved):

PR #465, opened 2021-07-15
Fix @everyone exploit

krofna:


algmyr (approved):
This is fine as at least as a band-aid for now. We can figure out a nicer solution (e.g. embeds or disallowing some mentions) later.

Issue #468, opened 2021-07-21
How to contribute?


iampjeetsingh:
I have implemented the following features in [my fork](https://github.com/iampjeetsingh/TLE) and would like to get them merged into the main branch.

- Codechef rating based roles
- CodeChef / AtCoder / Google Kickstart support in `;plot rating/performance` , `;ranklist` and `;handle pretty` using clist.by
- CodeChef and AtCoder support in `;handle identify` command
- Google Kickstart and LeetCode support in `;handle set` command (Todo: add them into identify)
- Ability to create HandleLists of users who are not present in the server.
- Updated `;remind` to [aryanc403/remind](https://github.com/aryanc403/remind)
- Ban feature (to restrict bot from responding to certain users)
- Commands to create roles for codeforces and codechef

How to proceed?

To try these features join [Priyansh31dec Server](https://discord.gg/AsBVjYScJH)

cheran-senthil:
Make partial PRs for each feature, and we can take a look. Some of these sound like good ideas and might get merged

Issue #94, opened 2019-07-27
atcoder reminder using clist api


krofna:

krofna:
https://clist.by/api/v1/doc/

PR #470, opened 2021-07-22
Codechef atcoder support

iampjeetsingh:
Implemented the following `;handle identify`, `;handle set`, `;handle list`, `;ranklist` commands for CodeChef / AtCoder / Google (Kickstart)


cheran-senthil:
Don't think we want these for other sites, general consensus is against this.

krofna:
#94 for reference

iampjeetsingh:
Which features were you talking in #468 then?

krofna:
@iampjeetsingh just ignore what cheran said. this requires detailed discussion. please provide your discord username so that we can invite you to the development server

iampjeetsingh:
My discord username is thesupremeone#1162

tle/util/clist_common.py
```diff
@@ -0,0 +1,287 @@
...
+            indexes.update(statistic['problems'].keys())
+    letters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'
+    index_map = {}
+    indexes = sorted(list(indexes))
+    for i, id in enumerate(indexes):
+        index_map[letters[i]] = id
+        indexes[i] = letters[i]
+    standings = clist.format_standings(statistics, index_map, indexes)
+    ranklist = CRanklist(contest, standings, deltas=deltas if rated else None, problems_indexes=indexes)
+    return ranklist
```
algmyr, line 262: A lot of these functions are on the form
```
if cf: minor logic + dispatch to cf_common

do complex clist logic
```
It might make sense to have a some overarching util (maybe `resource_util` for lack of better words) that does the backend agnostic work+the minor logic in the pseudocode above, and dispatches to cf_util or clist_util for specific implementations of e.g. ranklist. I.e. that ranklist functions does minor logic followed by dispatch to the specific impl of ranklist.

I suppose the resource code should also move there, just leaving the clist impl details here.

tle/util/db/user_db_conn.py
```diff
@@ -5,6 +5,7 @@
 from discord.ext import commands
```
algmyr, line 5: @meooow25 or @narutse, could you look at the db related code?

tle/util/scaper.py
```diff
@@ -0,0 +1,25 @@
+from bs4 import BeautifulSoup
+import requests
```
algmyr, line 2: Might want to make this use aiohttp like the rest of the code which makes the request async.

tle/util/scaper.py
```diff
@@ -0,0 +1,25 @@
+from bs4 import BeautifulSoup
```
algmyr, line 1: File comment:
If the core clist identify logic gets moved to clist_common this scraper code could be moved there.

algmyr (changes requested):

tle/util/clist_api.py
```diff
@@ -0,0 +1,271 @@
...
+                break
+        else:
+            objects = resp['objects']
+            results += objects
+            if(len(objects)<1000):
+                break
+        offset+=1000
+    return results
+
+class Contest(CfContest):
```
narutse, line 203: I think having Contest subclass CfContest to change one field is a bit weird. Should probably put this in a common class.

PR #471, opened 2021-07-27
Deactivate CSES

algmyr:
For reference, CSES now requires a login to check statistics. Let's deactivate all CSES support until there is a proper api for us to use.


dongliu0426:
> For reference, CSES now requires a login to check statistics. Let's deactivate all CSES support until there is a proper api for us to use.

Hi @algmyr , I made something that can check statistics on CSES (https://github.com/dongliu0426/cses-api, [example](http://api-cses.herokuapp.com/statistics/1068)). Can this be used for TLE?

PR #473, opened 2021-07-29
Tested and verified a method for pushing the bot onto heroku using docker.

shim1998:
### Benefits
- Doesn't require the addition of add ons or config variables
- Seems an alternative method of deploying the bot.
- Similar method can be used to deploy the bot on AWS or GCP machines.

Please mail me if you want proof or have doubts about how this works. 
Changes are welcomed in making this method more efficient.


iampjeetsingh:
What is the use of Procfile, heroku.yml, and changes you made in Dockerfile?

I mean running these commands in the root directory of the bot deploys it to Heroku without any issue
```
sudo heroku container:login
sudo heroku container:push worker -a <heroku app name>
sudo heroku container:release worker -a <heroku app name>
```

shim1998:
- docker-compose.yml file is there for a way to create and run the docker image locally via docker-compose engine. (It's optional to push that on your branch) 
- heroku.yml file was for testing out another method to deploy docker image on heroku, it works as well.
- However, DB fails to load at times because I am using free version of heroku so I have limited cpu and ram which results in sharding problems and restart the bot which results in losing the stored users on it.

fishy15:
Are you able to find a way to a way to keep the data even through restarts? I was working on a fork of TLE that would use the PostgreSQL database given for free on Heroku, but I have been lazy so I haven't finished testing to make sure everything is working right. However, if this fork isn't necessary, I don't see a point continuing to work on it.

cheran-senthil:
This is hard to merge with the new Dockerfile changes. I'm closing for this now, but you can reopen if you have a better deploy idea.

PR #475, opened 2021-08-21
Add median to ;plot speed

M7mdisk:
Added the option to use the median instead of the average for the `;plot speed` command by setting the `+median` flag.


algmyr (approved):
Nice addition, thanks!

PR #482, opened 2021-12-31
Added ;plot progress

### LINKED ISSUE #481: Feature Request: plot rating but contests evenly spread rather than by time
The x-axis would be contest# for that specific user. So if they compete in 5 contests, it would be from 1-5 evenly spreading out the contests. This could provide different more information, and a cleaner plot in some cases, especially for people that participate in contests in clumps and that have some inactivity. 

It would also give more interesting comparison between people. For multiple people specified in the command, the x-axis would go from 1 to max(numContest[user1], numContest[user2],..., numContest[user n-1]), and so you could see how you compare to X in the same number of contests. 

This could be in a command called ;plot progress.

ashwinbanwari:
Hello, this creates a new way to plot. It is similar to ;plot rating except the x-axis is contest number for that user (i.e. if a user compete in 10 contests, x-axis is from 1-10), thus evenly spacing out contests for them on the graph. For multi-user, the user with less contests has their graph cut off at contest x where x is that user's last contest. This provides an interesting way to compare users. 
image

RESOLVES #481

### DIFF
--- a/tle/cogs/graphs.py
+++ b/tle/cogs/graphs.py
@@ -39,45 +39,42 @@ def nice_sub_type(types):
                 'PRACTICE':'Practice: {}'}
     return [nice_map[t] for t in types]
 
-def _plot_rating(resp, mark='o'):
-    
-    for rating_changes in resp:
-        ratings, times = [], []
-        for rating_change in rating_changes:
-            ratings.append(rating_change.newRating)
-            times.append(dt.datetime.fromtimestamp(rating_change.ratingUpdateTimeSeconds))
-
-        plt.plot(times,
+def _plot_rating(plot_data, mark):
+    for ratings, when in plot_data:
+        plt.plot(when,
                  ratings,
                  linestyle='-',
                  marker=mark,
                  markersize=3,
                  markerfacecolor='white',
                  markeredgewidth=0.5)
-
     gc.plot_rating_bg(cf.RATED_RANKS)
+
+def _plot_rating_by_date(resp, mark='o'):
+    def gen_plot_data():
+        for rating_changes in resp:
+            ratings, times = [], []
+            for rating_change in rating_changes:
+                ratings.append(rating_change.newRating)
+                times.append(dt.datetime.fromtimestamp(rating_change.ratingUpdateTimeSeconds))
+            yield (ratings, times)
+
+    _plot_rating(gen_plot_data(), mark)
     plt.gcf().autofmt_xdate()
 
-def _plot_progress(resp, mark='o'):
-    
-    for rating_changes in resp:
-        ratings, indices = [], []
-        index = 1
-        for rating_change in rating_changes:
-            ratings.append(rating_change.newRating)
-            indices.append(index)
-            index += 1
+def _plot_rating_by_contest(resp, mark='o'):
+    def gen_plot_data():
+        for rating_changes in resp:
+            ratings, indices = [], []
+            index = 1
+            for rating_change in rating_changes:
+                ratings.append(rating_change.newRating)
+                indices.append(index)
+                index += 1
+            yield (ratings, indices)
 
-        plt.plot(indices,
-                 ratings,
-                 linestyle='-',
-                 marker=mark,
-                 markersize=3,
-                 markerfacecolor='white',
-                 markeredgewidth=0.5)
+    _plot_rating(gen_plot_data(), mark)
 
-    gc.plot_rating_bg(cf.RATED_RANKS)
-    plt.gcf().autofmt_xdate()
 
 def _classify_submissions(submissions):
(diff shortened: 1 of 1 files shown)


krofna:
@algmyr what do you think? should this just be a flag for plot rating or this is good?

algmyr:
> @algmyr what do you think? should this just be a flag for plot rating or this is good?

Considering the logic in the command function is pretty much identical it probably makes more sense as a flag.

ashwinbanwari:
> > @algmyr what do you think? should this just be a flag for plot rating or this is good?
> 
> 
> 
> Considering the logic in the command function is pretty much identical it probably makes more sense as a flag.

I can change this to a flag. I noticed while testing, putting the legend in the bottom right was much better on average than putting it in top left (since rating graphs are often more like logarithmic curve than linear), can I change the plot in general to make it bottom right legend or should I do it just for this flag (where this problem is more severe)

tle/cogs/graphs.py
```diff
@@ -58,6 +58,27 @@ def _plot_rating(resp, mark='o'):
     gc.plot_rating_bg(cf.RATED_RANKS)
     plt.gcf().autofmt_xdate()
 
+def _plot_progress(resp, mark='o'):
```
algmyr, line 61: I would prefer some deduplication of logic here. E.g. we can have something like
```
def _plot_rating(plot_data, mark):
    for when, ratings in plot_data:
        plt.plot(when,
                 ratings,
                 linestyle='-',
                 marker=mark,
                 markersize=3,
                 markerfacecolor='white',
                 markeredgewidth=0.5)
    gc.plot_rating_bg(cf.RATED_RANKS)

def _plot_rating_by_date(resp, mark='o'):
    def gen_plot_data():
        for rating_changes in resp:
            ratings, times = [], []
            for rating_change in rating_changes:
                ratings.append(rating_change.newRating)
                times.append(dt.datetime.fromtimestamp(rating_change.ratingUpdateTimeSeconds))
            yield (rating, times)

    _plot_rating(gen_plot_data(), mark)
    plt.gcf().autofmt_xdate()

def _plot_rating_by_contest(...):
    ... similar to above, minus the date formatting line
```
which factors out the common plot formatting.

tle/cogs/graphs.py
```diff
@@ -265,6 +286,63 @@ def max_prefix(user):
         discord_common.set_author_footer(embed, ctx.author)
         await ctx.send(embed=embed, file=discord_file)
 
+    @plot.command(brief='Plot Codeforces rating graph by contests', usage='[+zoom] [+peak] [handles...] [d>=[[dd]mm]yyyy] [d<[[dd]mm]yyyy]')
```
algmyr, line 289: As discussed, considering almost all logic is the same, make this a flag for the rating command instead.

algmyr (changes requested):

algmyr:
> I noticed while testing, putting the legend in the bottom right was much better on average than putting it in top left (since rating graphs are often more like logarithmic curve than linear), can I change the plot in general to make it bottom right legend or should I do it just for this flag (where this problem is more severe)

Getting the positioning correct is not easy. Bottom-right will mess up the plot for people who have a big rating dip or who converge to a rating lower than the starting one. Top-left was chosen because it mostly only messes up for people who never exceed their starting rating. It's not perfect, but it tended to be better behaved than letting matplotlib pick the legend position.

I would expect curves to often be similar between the date and num_contest plots, so it probably makes sense to have the same behavior for both. I would vote top-left for both, but @meooow25 @krofna please chime in.

ashwinbanwari:
> > I noticed while testing, putting the legend in the bottom right was much better on average than putting it in top left (since rating graphs are often more like logarithmic curve than linear), can I change the plot in general to make it bottom right legend or should I do it just for this flag (where this problem is more severe)
> 
> 
> 
> Getting the positioning correct is not easy. Bottom-right will mess up the plot for people who have a big rating dip or who converge to a rating lower than the starting one. Top-left was chosen because it mostly only messes up for people who never exceed their starting rating. It's not perfect, but it tended to be better behaved than letting matplotlib pick the legend position.
> 
> 
> 
> I would expect curves to often be similar between the date and num_contest plots, so it probably makes sense to have the same behavior for both. I would vote top-left for both, but @meooow25 @krofna please chime in.

hmm I think I may not have observed bottom right messing up that much because of the change that everyone starts with 0 rating (and few ever stay around there), but I can see how that could be problem of older users rating graphs. Anyway, I'll wait for other opinions.

ashwinbanwari:
For now, I made it a flag (left the legend stuff untouched). maybe we can get this merged in the mean time? @algmyr

algmyr (approved):
LGTM

PR #490, opened 2022-04-24
Fix handle already associated check in identify

krofna:


tle/util/db/user_db_conn.py
```diff
@@ -396,11 +396,11 @@ def get_handle(self, user_id, guild_id):
         res = self.conn.execute(query, (user_id, guild_id)).fetchone()
         return res[0] if res else None
 
-    def get_user_id(self, handle, guild_id):
+    def get_user_id(self, handle, guild_id, active=1):
```
algmyr, line 399: Let's make the python api nicer and accept a `bool`? And then pass `int(active)` to the query

krofna: i do >= so int is more fitting than bool

algmyr (approved):
LGTM

PR #491, opened 2022-04-24
Allow ;handle remove by cf handle

krofna:


algmyr (approved):
LGTM, merge after testing it

PR #498, opened 2022-05-03
Develop duelnotags

DenjellBoone:
Added the possibility to ban tags for all commands that support tags.
Prefix is ~.

### DIFF
--- a/tle/cogs/codeforces.py
+++ b/tle/cogs/codeforces.py
@@ -177,7 +177,7 @@ class Codeforces(commands.Cog):
         Add tags with "+" before them.
         Exclude tags with "~" before them.
         """
-        handles = [arg for arg in args if arg[0] != '+']
+        handles = [arg for arg in args if arg[0] != '+' and arg[0] != '~']
         tags = cf_common.parse_tags(args)
         notags = cf_common.parse_tags(args, '~')
 
--- a/tle/cogs/codeforces.py
+++ b/tle/cogs/codeforces.py
@@ -177,7 +177,7 @@ class Codeforces(commands.Cog):
         Add tags with "+" before them.
         Ban tags with "~" before them.
         """
-        handles = [arg for arg in args if arg[0] != '+' and arg[0] != '~']
+        handles = [arg for arg in args if arg[0] not in '+~']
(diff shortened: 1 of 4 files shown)


tle/cogs/codeforces.py
```diff
@@ -169,13 +171,15 @@ def make_page(chunk):
...
 
-    @commands.command(brief='Create a mashup', usage='[handles] [+tags]')
+    @commands.command(brief='Create a mashup', usage='[handles] [+tags] [~tags]')
     async def mashup(self, ctx, *args):
         """Create a mashup contest using problems within +-100 of average rating of handles provided.
         Add tags with "+" before them.
+        Ban tags with "~" before them.
         """
-        handles = [arg for arg in args if arg[0] != '+']
+        handles = [arg for arg in args if arg[0] != '+' and arg[0] != '~']
```
krofna, line 180: arg[0] not in '+~'

krofna (commented):
seems good apart from that nit

krofna (commented):
if len(bantags) > 0 make it unofficial

DenjellBoone:
I made the changes. Check again

krofna (approved):

tle/util/codeforces_common.py
```diff
@@ -281,8 +281,8 @@ def parse_date(arg):
         raise ParamParseError(f'{arg} is an invalid date argument')
 
 
-def parse_tags(args):
-    tags = [x[1:] for x in args if x[0] == '+']
+def parse_tags(args, identifier = '+'):
```
algmyr, line 284: Would it be better to drop the default? It would make the call sites more explicit.

For clarity it might also be nice to have a better argument name and even making it keyword only, i.e.

```
def parse_tags(args, *, prefix):
    ...
    
tags = parse_tags(args, prefix='+')
ban_tags = parse_tags(args, prefix='~')
```

DenjellBoone: I added the changes. Please check again.

algmyr (approved):

algmyr:
LGTM, I'm assuming you found all the calls and updated them, I guess we'll find that out soon enough. :P

PR #500, opened 2022-05-04
Update python version specs. Drop 3.7 support.

algmyr:
Recent scipy sets an upper bound on the python version, which prevented
updating. It also bumped the lower bound to 3.8. I think it should be
fine ditching 3.7 support to keep up with other packages.


cheran-senthil (approved):

algmyr:
Will add a new PR in a bit.

Issue #244, opened 2019-12-15
option to show only official participants in ranklist


krofna:
Note that this shouldn't be default, since people primarily use ;ranklist for virtual contests.
So I suggest -virtual (as opposed to +virtual) option.

![image](https://user-images.githubusercontent.com/1102695/70863604-2363b780-1f4a-11ea-93ae-84a2983f670d.png)

algmyr:
Likely solved by #509, open again if not.

Issue #72, opened 2019-07-08
teams aren't shown in ;ranklist


krofna:

algmyr:
Likely solved by #509, open again if not.

PR #509, opened 2023-04-30
Add support for +official flag in ranklist

prabh1601:
1. Ranklist can now show official rated participants in ranklist (for eg : show contestants with < 2100 rating in a div2/hide virtual contestants in a div1)
2. Ranklist will fallback to predicted rating changes if fetched rating changes are not available
3. Remove predicted rating changes for events with "icpc" in the name (these are supposed to represent icpc mirrors or optimization contests)
4. Minor refactorings

### DIFF
--- a/tle/cogs/contests.py
+++ b/tle/cogs/contests.py
@@ -460,7 +460,7 @@ class Contests(commands.Cog):
             if contest.phase == 'BEFORE':
                 raise ContestCogError(f'Contest `{contest.id} | {contest.name}` has not started')
             ranklist = await cf_common.cache2.ranklist_cache.generate_ranklist(contest.id, fetch_changes=True,
-                                                                               show_unofficial=show_official is False)
+                                                                               show_unofficial=not show_official)
 
         await wait_msg.delete()
         await ctx.channel.send(embed=self._make_contest_embed_for_ranklist(ranklist))
@@ -479,7 +479,7 @@ class Contests(commands.Cog):
                 continue
 
             # Database has correct handle ignoring case, update to it
-            handle = standing.party.teamName or standing.party.members[0].handle
+            handle = rl.Ranklist.get_ranklist_lookup_key(standing)
             if vc and standing.party.participantType != 'VIRTUAL':
                 continue
             handle_standings.append((handle, standing))
--- a/tle/cogs/contests.py
+++ b/tle/cogs/contests.py
@@ -442,21 +442,26 @@ class Contests(commands.Cog):
         return embed
 
     @commands.command(brief='Show ranklist for given handles and/or server members')
-    async def ranklist(self, ctx, contest_id: int, *handles: str):
-        """Shows ranklist for the contest with given contest id. If handles contains
-        '+server', all server members are included. No handles defaults to '+server'.
+    async def ranklist(self, ctx, contest_id: int, *args: str):
         """
-        handles = await cf_common.resolve_handles(ctx, self.member_converter, handles, maxcnt=None, default_to_all_server=True)
+        Shows ranklist for the contest with given contest id. If handles contains
+        '+server', all server members are included. No handles defaults to '+server'.
+        Use '+official' for only showing rated participants for the round
+        """
+        (show_official,), handles = cf_common.filter_flags(args, ['+official'])
+        handles = await cf_common.resolve_handles(ctx, self.member_converter, handles, maxcnt=None,
+                                                  default_to_all_server=True)
         contest = cf_common.cache2.contest_cache.get_contest(contest_id)
         wait_msg = await ctx.channel.send('Generating ranklist, please wait...')
         ranklist = None
         try:
-            ranklist = cf_common.cache2.ranklist_cache.get_ranklist(contest)
+            ranklist = cf_common.cache2.ranklist_cache.get_ranklist(contest, show_official)
         except cache_system2.RanklistNotMonitored:
             if contest.phase == 'BEFORE':
                 raise ContestCogError(f'Contest `{contest.id} | {contest.name}` has not started')
-            ranklist = await cf_common.cache2.ranklist_cache.generate_ranklist(contest.id,
-                                                                            fetch_changes=True)
+            ranklist = await cf_common.cache2.ranklist_cache.generate_ranklist(contest.id, fetch_changes=True,
+                                                                               show_unofficial=show_official is False)
+
         await wait_msg.delete()
         await ctx.channel.send(embed=self._make_contest_embed_for_ranklist(ranklist))
         await self._show_ranklist(channel=ctx.channel, contest_id=contest_id, handles=handles, ranklist=ranklist)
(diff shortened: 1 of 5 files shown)


prabh1601:
Possibly solves #72 and #244

tle/util/cache_system2.py
```diff
@@ -579,51 +583,83 @@ async def _monitor_task(self, _):
...
+        ranklist = None
+        if fetch_changes:
+            ranklist = await self._get_ranklist_with_fetched_changes(contest_id, show_unofficial)
+        if ranklist is None:
+            # Either predict_changes was true or fetching rating changes failed
+            ranklist = await self._get_ranklist_with_predicted_changes(contest_id, show_unofficial)
+
+        # for some reason Educational contests also have div1 peeps in the official standings.
+        # hence we need to manually weed them out
+        if not show_unofficial and 'Educational' in ranklist.contest.name:
```
algmyr, line 661: This shouldn't be `and`, should it?

tle/util/ranklist/ranklist.py
```diff
@@ -33,20 +34,37 @@ def __init__(self, contest, problems, standings, fetch_time, *, is_rated):
...
-                # Apparently ghosts don't have team ID.
-                id_ = row.party.teamName
-            else:
-                id_ = row.party.teamId or row.party.members[0].handle
+            id_ = row.party.teamName or row.party.members[0].handle
             self.standing_by_id[id_] = row
 
-        self.delta_by_handle = None
-        self.deltas_status = None
+    def fix_rated_standings(self):
```
algmyr, line 49: Make this more descriptive. Something with `remove_unofficial` would make sense

tle/util/ranklist/ranklist.py
```diff
@@ -33,20 +34,37 @@ def __init__(self, contest, problems, standings, fetch_time, *, is_rated):
...
 
+    def _create_inverse_standings(self):
         self.standing_by_id = HandleDict()
         for row in self.standings:
-            if row.party.ghost:
-                # Apparently ghosts don't have team ID.
-                id_ = row.party.teamName
-            else:
-                id_ = row.party.teamId or row.party.members[0].handle
+            id_ = row.party.teamName or row.party.members[0].handle
```
algmyr, line 46: What's the rationale for this change in logic?

tle/util/ranklist/ranklist.py
```diff
@@ -33,20 +34,37 @@ def __init__(self, contest, problems, standings, fetch_time, *, is_rated):
...
 
-        self.delta_by_handle = None
-        self.deltas_status = None
+    def fix_rated_standings(self):
+        if self.delta_by_handle is None:
+            raise DeltasNotPresentError
+
+        fixed_standings = []
+        current_rated_rank = 1
+        last_rated_rank, last_rated_score = 0, (-1, -1)
```
algmyr, line 55: Split this into two assignments.

tle/util/ranklist/ranklist.py
```diff
@@ -33,20 +34,37 @@ def __init__(self, contest, problems, standings, fetch_time, *, is_rated):
...
+            id_ = row.party.teamName or row.party.members[0].handle
             self.standing_by_id[id_] = row
 
-        self.delta_by_handle = None
-        self.deltas_status = None
+    def fix_rated_standings(self):
+        if self.delta_by_handle is None:
+            raise DeltasNotPresentError
+
+        fixed_standings = []
```
algmyr, line 53: with name change above, maybe this should be `official_standings`

tle/util/ranklist/ranklist.py
```diff
@@ -33,20 +34,37 @@ def __init__(self, contest, problems, standings, fetch_time, *, is_rated):
...
+        current_rated_rank = 1
+        last_rated_rank, last_rated_score = 0, (-1, -1)
+        for contestant in self.standings:
+            handle = contestant.party.teamName or contestant.party.members[0].handle
```
algmyr, line 57: It seems this logic is repeated a bunch, maybe it should be broken out into a function?

algmyr (changes requested):

tle/cogs/contests.py
```diff
@@ -442,21 +442,26 @@ def _make_contest_embed_for_vc_ranklist(ranklist, vc_start_time=None, vc_end_tim
...
         try:
-            ranklist = cf_common.cache2.ranklist_cache.get_ranklist(contest)
+            ranklist = cf_common.cache2.ranklist_cache.get_ranklist(contest, show_official)
         except cache_system2.RanklistNotMonitored:
             if contest.phase == 'BEFORE':
                 raise ContestCogError(f'Contest `{contest.id} | {contest.name}` has not started')
-            ranklist = await cf_common.cache2.ranklist_cache.generate_ranklist(contest.id,
-                                                                            fetch_changes=True)
+            ranklist = await cf_common.cache2.ranklist_cache.generate_ranklist(contest.id, fetch_changes=True,
+                                                                               show_unofficial=show_official is False)
```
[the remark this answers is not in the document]
prabh1601: Oh yes, this is good.

tle/util/cache_system2.py
```diff
@@ -579,51 +583,83 @@ async def _monitor_task(self, _):
...
+        try:
+            changes = await cf.contest.ratingChanges(contest_id=contest_id)
+            # For contests intended to be rated but declared unrated, an empty list is returned.
+            is_rated = len(changes) > 0
+        except cf.RatingChangesUnavailableError:
+            pass
+
+        ranklist = None
+        if is_rated:
             ranklist = Ranklist(contest, problems, standings, now, is_rated=is_rated)
```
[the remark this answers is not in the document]
prabh1601: Yes, this is intended in the change.

The previous code did rank list calculations either with predict_changes or fetch_changes as true. However, this model fails to work in a few cases. Recent/Ongoing contests were fetched from ranklistcache rather than an API call in the previous code version. 

The new version works as follows: In order to fetch the official rank list (with +official), the cache is not used, rather we make an API call, which will, in turn, won't return rating changes(it's a recent/ongoing contest). This iteration of code helps to fallback to predict_changes in case fetching rating changes fails (is_rated being false and in turn returning none as a rank list).

This would also result in showing predicted deltas in the case of old unrated contests as a side benefit.

tle/util/cache_system2.py
prabh1601: I think it should be `and`. This line aims to remove unofficial people when the official educational contest rank list also returns div1s, which has the following conditions
1. We passed +official flag to command -> not show_unofficial
2. It was an Educational Contest

tle/util/ranklist/ranklist.py
prabh1601: Oh very true, I agree this is a much better choice for the naming conventions.

tle/util/ranklist/ranklist.py
prabh1601: This will solve #72, currently, the rank list uses teamID (which is an integer - [API Docs](https://codeforces.com/apiHelp/objects#Party)) to store the non-ghost teams. This leads to the fact that a lookup cant be performed based on the team name. We can exclusively use team names for teams other than Ghosts as well

With this change, now the lookup is based on 
1. team-name in case of a team 
2. first team member's handle in case it's not a team (in which case the team name would be None)

tle/util/ranklist/ranklist.py
prabh1601: Agreed. Much better convention

tle/util/ranklist/ranklist.py
prabh1601: I agree, will make life a lot easier.

tle/util/ranklist/ranklist.py
algmyr: Gotcha

algmyr (approved):
Cool, thanks for the PR!

PR #519, opened 2024-01-20
Add a way to install files using Pip

### LINKED ISSUE #518: Installation is broken on some systems
Following the installation instructions (both with and without Docker) can lead to errors on some systems, including the Raspberry Pi 4. This is caused by Poetry not searching the correct sources for some dependencies (see https://github.com/python-poetry/poetry/issues/4854).

This can be resolved by using Pip to install the requirements instead of Poetry. However, for users unfamiliar with Poetry, this is not intuitive.

E-M-Bailey:
fixes #518 

This adds a new run script, run-pip.sh, which uses Pip instead of Poetry to install dependencies. It also modifies the README with appropriate instructions.

### DIFF
--- a/run-pip.sh
+++ b/run-pip.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# Get to a predictable directory, the directory of this script
+cd "$(dirname "$0")"
+
+[ -e environment ] && . ./environment
+python3 -m venv .venv
+. .venv/bin/activate
+
+while true; do
+    git pull
+    poetry export --without-hashes > requirements.txt
+    python -m pip install --requirement requirements.txt
+    FONTCONFIG_FILE=$PWD/extra/fonts.conf python -m tle
+
+    (( $? != 42 )) && break
+
+    echo '==================================================================='
+    echo '=                       Restarting                                ='
+    echo '==================================================================='
+done
--- a/README.md
+++ b/README.md
@@ -49,6 +49,8 @@ python3.7 -m pip install poetry
 python3.7 -m poetry install
 ```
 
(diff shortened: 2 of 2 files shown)


run-pip.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+# Get to a predictable directory, the directory of this script
+cd "$(dirname "$0")"
+
+[ -e environment ] && . ./environment
+python3 -m venv .venv
+. .venv/bin/activate
```
algmyr, line 8: WDYT about making this flexible wrt venv usage? So, add something like this to `environment.template`
```
# Set to enable automatic venv usage.
export VENV_DIR=""
```
and then
```suggestion
if [[ -n "${VENV_DIR}" ]]; then
    python3 -m venv "${VENV_DIR}"
    . "${VENV_DIR}/bin/activate"
fi
```
adding the same to the regular startup script might also be a nice idea.

And of course, then document this feature in some sensible place in the readme.

algmyr (changes requested):
Hi, sorry for not getting to this for a bit. Added a comment about making the venv usage (which I like) common among the script and configurable. Should be a pretty minor change, I hope you don't mind. :)

algmyr:
I guess I could make that change myself in a follow-up.

run-pip.sh
algmyr: Follow-up...

E-M-Bailey:
Sorry, I missed your original response. I'm happy to try making the changes myself sometime this weekend, although I don't have a good way of testing the non-Pip version.

algmyr:
> Sorry, I missed your original response. I'm happy to try making the changes myself sometime this weekend, although I don't have a good way of testing the non-Pip version.

No worries, I made the changes already in #520. You'll have to make a change to add an environment variable but that should be about it. :)

PR #520, opened 2024-01-26
Auto venv follow-up changes.

algmyr:
Some convenience stuff for venv users.


algmyr:
@E-M-Bailey FYI I made some changes related to the venv usage which removes the `.venv` default. Adjust your `environment` to list your venv there as per the readme and things should just work.

PR #526, opened 2025-04-05
Require python 3.9 (oldest non-EOL), update dependencies a bit

algmyr:
I left discord.py at pre 2.0 because that has possible breaking changes. This update can be done later.


cheran-senthil (approved):
Can we move off of poetry yet or is that something for later?

algmyr:
I would stay in poetry for the time being. If someone would like to switch to uv or whatever in the future that wouldn't be a terrible idea.

PR #528, opened 2025-04-05
Sort all imports, fix some outdated typing stuff, add ruff config

algmyr:
Just cleaning up some stuff without being overly invasive. I'll see if formatting the whole repo makes sense later on...

### DIFF
--- a/tle/util/codeforces_api.py
+++ b/tle/util/codeforces_api.py
@@ -1,19 +1,16 @@
 import asyncio
 from collections import defaultdict
 from collections import deque
+from collections.abc import Iterable
+from collections.abc import Iterator
+from collections.abc import Sequence
 import functools
 import itertools
 import logging
 import time
 from typing import Any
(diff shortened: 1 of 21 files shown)


tle/util/codeforces_api.py
```diff
@@ -146,7 +153,7 @@ class Member(NamedTuple):
 class Party(NamedTuple):
     """Codeforces party."""
     contestId: Optional[int]
-    members: List[Member]
+    members: list[Member]
```
cheran-senthil, line 156: They renamed `List` to `list` huh

cheran-senthil (approved):
lgtm, can probably update the bot after this to verify everything works

tle/util/codeforces_api.py
algmyr: More like added subscripting to list/dict/... which removes the need for `typing.List` and the like

PR #530, opened 2025-04-20
trusted role init

farmersrice:


tle/cogs/handles.py
```diff
@@ -329,13 +330,54 @@ async def update_member_rank_role(member, role_to_assign, *, reason):
...
+                    # Check if rating reached 1900+ before the cutoff date
+                    if change.newRating >= 1900 and change.ratingUpdateTimeSeconds < cutoff_timestamp:
+                        is_trusted_candidate = True
+                        break # Found one instance, no need to check further
+
+                if is_trusted_candidate:
+                        try:
+                            await member.add_roles(trusted_role, reason='Historical rating >= 1900 before Aug 2024')
+                        except discord.Forbidden:
+                             print(f"WARN: Missing permissions to add Trusted role to {member.display_name} in {member.guild.name}")
```
algmyr, line 364: Actually use `logging` to log in all of these places.

tle/cogs/handles.py
```diff
@@ -329,13 +330,54 @@ async def update_member_rank_role(member, role_to_assign, *, reason):
...
         if to_remove:
             await member.remove_roles(*to_remove, reason=reason)
+
+        # --- Grant 'Trusted' Role based on historical rating ---
+        handle = cf_common.user_db.get_handle(member.id, member.guild.id)
+        trusted_role = discord.utils.get(member.guild.roles, name='Trusted')
+        if handle and trusted_role and trusted_role not in member.roles: # only check if not already trusted
+            try:
+                # Define cutoff date (August 1, 2024 00:00 UTC)
+                cutoff_timestamp = dt.datetime(2024, 8, 1, tzinfo=dt.timezone.utc).timestamp()
```
algmyr, line 350: Why this magic date? Add a comment. Also the current comment is useless and adds nothing.

tle/cogs/handles.py
```diff
@@ -329,13 +330,54 @@ async def update_member_rank_role(member, role_to_assign, *, reason):
...
                 role_names_to_remove.add('Purgatory')
         to_remove = [role for role in member.roles if role.name in role_names_to_remove]
         if to_remove:
             await member.remove_roles(*to_remove, reason=reason)
+
+        # --- Grant 'Trusted' Role based on historical rating ---
+        handle = cf_common.user_db.get_handle(member.id, member.guild.id)
+        trusted_role = discord.utils.get(member.guild.roles, name='Trusted')
+        if handle and trusted_role and trusted_role not in member.roles: # only check if not already trusted
+            try:
```
algmyr, line 348: This big try block catching stuff from all over sucks, please limit the CF related exceptions to where the error can actually happen.

tle/cogs/handles.py
```diff
@@ -329,13 +330,54 @@ async def update_member_rank_role(member, role_to_assign, *, reason):
...
+                        is_trusted_candidate = True
+                        break # Found one instance, no need to check further
+
+                if is_trusted_candidate:
+                        try:
+                            await member.add_roles(trusted_role, reason='Historical rating >= 1900 before Aug 2024')
+                        except discord.Forbidden:
+                             print(f"WARN: Missing permissions to add Trusted role to {member.display_name} in {member.guild.name}")
+                        except discord.HTTPException as e:
+                             print(f"WARN: Failed to add Trusted role to {member.display_name} in {member.guild.name}: {e}")
```
algmyr, line 366: Fix this inconsistent indent

tle/cogs/handles.py
```diff
@@ -329,13 +330,54 @@ async def update_member_rank_role(member, role_to_assign, *, reason):
         on the member will be removed.
         """
         role_names_to_remove = {rank.title for rank in cf.RATED_RANKS}
+        # Ensure 'Trusted' is never removed by this process
+        role_names_to_remove.discard('Trusted')
```
algmyr, line 334: Do you expect mike to add a Trusted rank on CF?

tle/cogs/handles.py
```diff
@@ -329,13 +330,54 @@ async def update_member_rank_role(member, role_to_assign, *, reason):
...
+
         if role_to_assign is not None:
             role_names_to_remove.discard(role_to_assign.name)
             if role_to_assign.name not in ['Newbie', 'Pupil', 'Specialist', 'Expert']:
                 role_names_to_remove.add('Purgatory')
         to_remove = [role for role in member.roles if role.name in role_names_to_remove]
         if to_remove:
             await member.remove_roles(*to_remove, reason=reason)
+
```
algmyr, line 343: The logic here has nothing to do with the rank update, make it a separate function and call it separately where relevant. `maybe_add_trusted_role` or whatever is a fitting name.

tle/cogs/handles.py
```diff
@@ -329,13 +330,54 @@ async def update_member_rank_role(member, role_to_assign, *, reason):
...
+        # --- Grant 'Trusted' Role based on historical rating ---
+        handle = cf_common.user_db.get_handle(member.id, member.guild.id)
+        trusted_role = discord.utils.get(member.guild.roles, name='Trusted')
+        if handle and trusted_role and trusted_role not in member.roles: # only check if not already trusted
+            try:
+                # Define cutoff date (August 1, 2024 00:00 UTC)
+                cutoff_timestamp = dt.datetime(2024, 8, 1, tzinfo=dt.timezone.utc).timestamp()
+                rating_changes = await cf.user.rating(handle=handle)
```
algmyr, line 351: This is the thing that can fail with cf exceptions, so try-except *this* with the CF related exceptions

tle/cogs/handles.py
```diff
@@ -329,13 +330,54 @@ async def update_member_rank_role(member, role_to_assign, *, reason):
...
+
+                is_trusted_candidate = False
+                for change in rating_changes:
+                    # Check if rating reached 1900+ before the cutoff date
+                    if change.newRating >= 1900 and change.ratingUpdateTimeSeconds < cutoff_timestamp:
+                        is_trusted_candidate = True
+                        break # Found one instance, no need to check further
```
algmyr, line 358: This is just
```
is_trusted_candidate = any(
    change.newRating >= 1900 and change.ratingUpdateTimeSeconds < cutoff_timestamp
    for change in rating_changes
)
```

tle/cogs/handles.py
```diff
@@ -329,13 +330,54 @@ async def update_member_rank_role(member, role_to_assign, *, reason):
...
+
         if role_to_assign is not None:
             role_names_to_remove.discard(role_to_assign.name)
             if role_to_assign.name not in ['Newbie', 'Pupil', 'Specialist', 'Expert']:
                 role_names_to_remove.add('Purgatory')
         to_remove = [role for role in member.roles if role.name in role_names_to_remove]
         if to_remove:
             await member.remove_roles(*to_remove, reason=reason)
+
+        # --- Grant 'Trusted' Role based on historical rating ---
```
algmyr, line 344: Again, separate function. "Grant 'Trusted' Role based on historical rating" would be a sensible docstring for such a function.

tle/cogs/handles.py
```diff
@@ -329,13 +330,54 @@ async def update_member_rank_role(member, role_to_assign, *, reason):
...
+            except cf.NotFoundError:
+                # User rating info not found via API, ignore for trusted check
+                print(f"INFO: Rating history not found for handle {handle} during trusted check.")
+            except cf.CodeforcesApiError as e:
+                # Log API errors appropriately in a real scenario
+                print(f"WARN: API Error fetching rating for {handle} during trusted check: {e}")
+            except Exception as e:
+                # Log other unexpected errors appropriately in a real scenario
+                print(f"ERROR: Unexpected error checking trusted status for {member.display_name} ({handle}): {e}")
+        elif not trusted_role:
```
algmyr, line 377: In the factored out function doing only the one thing, please do this check immediately at the top and early exit. Having this checked at the end doesn't make much sense.

tle/cogs/handles.py
```diff
@@ -329,13 +330,54 @@ async def update_member_rank_role(member, role_to_assign, *, reason):
...
             if role_to_assign.name not in ['Newbie', 'Pupil', 'Specialist', 'Expert']:
                 role_names_to_remove.add('Purgatory')
         to_remove = [role for role in member.roles if role.name in role_names_to_remove]
         if to_remove:
             await member.remove_roles(*to_remove, reason=reason)
+
+        # --- Grant 'Trusted' Role based on historical rating ---
+        handle = cf_common.user_db.get_handle(member.id, member.guild.id)
+        trusted_role = discord.utils.get(member.guild.roles, name='Trusted')
+        if handle and trusted_role and trusted_role not in member.roles: # only check if not already trusted
```
algmyr, line 347: * Check `handle` early, and early exit if it is missing
* Check `trusted_role` early, and early exit if it is missing
* Negate the check and early exit if the trusted role is already set

tle/cogs/handles.py
```diff
@@ -830,6 +872,54 @@ async def role(self, ctx, action: str, which: str):
...
+        # Find the Purgatory role
+        purgatory_role = discord.utils.get(guild.roles, name=purgatory_role_name)
+        if purgatory_role is None:
+            # This case might indicate a server setup issue, but we proceed as if the user is not in purgatory
+            self.logger.warning(f"Role '{purgatory_role_name}' not found in guild {guild.name} ({guild.id}).")
+        elif purgatory_role in target_user.roles:
+            await ctx.send(embed=discord_common.embed_alert(
+                f"Cannot grant Trusted role to {target_user.mention}. User is currently in Purgatory."
+            ))
+            return
```
algmyr, line 899: What's the long term plan? Getting rid of the purgatory role and use trusted instead? Having both feels weird

tle/cogs/handles.py
```diff
@@ -830,6 +872,54 @@ async def role(self, ctx, action: str, which: str):
...
+             return
+
+        # Grant the Trusted role
+        try:
+            await target_user.add_roles(trusted_role, reason=f"Referred by {ctx.author.name} ({ctx.author.id})")
+            await ctx.send(
+                f"{trusted_role.mention} role granted to {target_user.mention} by {ctx.author.mention}."
+            )
+        except discord.Forbidden:
+            raise HandleCogError(f"I lack the permissions to assign the '{trusted_role_name}' role.")
```
algmyr, line 920: Why `I`? Just describe the failure. "Failed to assign the ... role due to a permission error." or whatever.

tle/cogs/handles.py
```diff
@@ -830,6 +872,54 @@ async def role(self, ctx, action: str, which: str):
...
+            raise HandleCogError(f"The role '{trusted_role_name}' does not exist in this server.")
+
+        # Check if target user already has the role
+        if trusted_role in target_user.roles:
+             await ctx.send(embed=discord_common.embed_neutral(
+                f"{target_user.mention} already has the {trusted_role.mention} role."
+            ))
+             return
+
+        # Grant the Trusted role
```
algmyr, line 913: I see no checks on the calling user. So anyone can refer anyone that isn't in purgatory? Feels like you're missing some checks here.

algmyr (changes requested):

tle/cogs/handles.py
algmyr: Nvm, it's part of the decorator

tle/cogs/handles.py
farmersrice: have both roles

tle/cogs/handles.py
farmersrice: We have to check the current rating too, and it will be called in all times where we add it and no other times. I moved it separately but it will still be invoked here

tle/cogs/handles.py
farmersrice: otherwise we need to make another call to get the user's rating at the time of being referred, as opposed to just checking the purgatory status to see if eligible for trusted

farmersrice:
the current commands are tested and working, will add the grandfather command next

tle/cogs/handles.py
```diff
@@ -830,6 +865,145 @@ async def role(self, ctx, action: str, which: str):
...
+            member_joined_at_aware = member.joined_at.replace(tzinfo=dt.timezone.utc)
+
+            if member_joined_at_aware >= cutoff_date:
+                skipped_join_date += 1
+                continue
+
+            # Check if user already has the Trusted role
+            if trusted_role in member.roles:
+                skipped_already_trusted += 1
+                continue
```
algmyr, line 972: Again, don't describe the what but the why. E.g. like above
```suggestion
            if trusted_role in member.roles:
                # User already trusted, skip
                skipped_already_trusted += 1
                continue
```

tle/cogs/handles.py
```diff
@@ -830,6 +865,145 @@ async def role(self, ctx, action: str, which: str):
...
+
+            # Make member.joined_at timezone-aware (assuming it's UTC, which discord.py uses)
+            member_joined_at_aware = member.joined_at.replace(tzinfo=dt.timezone.utc)
+
+            if member_joined_at_aware >= cutoff_date:
+                skipped_join_date += 1
+                continue
+
+            # Check if user already has the Trusted role
+            if trusted_role in member.roles:
```
algmyr, line 970: Can we move this earlier? This feels like the first thing to check. Ordering like this makes more sense to me
* No need to process, skip
* Ineligible because of X, skip
* Ineligible because of Y, skip
* ...
* Eligible, do the thing

tle/cogs/handles.py
```diff
@@ -830,6 +865,145 @@ async def role(self, ctx, action: str, which: str):
...
+                skipped_already_trusted += 1
+                continue
+
+            # Check if user has the Purgatory role (only if the role exists)
+            if purgatory_role is not None and purgatory_role in member.roles:
+                skipped_purgatory += 1
+                continue
```
algmyr, line 977:
```suggestion
            if purgatory_role is not None and purgatory_role in member.roles:
                # User has purgatory role so is not eligible, skip
                skipped_purgatory += 1
                continue
```

algmyr (changes requested):
Probably the last batch of comments. If people are fine with the political side of this then I'm fine with merging this after the changes

tle/cogs/handles.py
```diff
@@ -830,6 +865,145 @@ async def role(self, ctx, action: str, which: str):
...
+        status_message = await ctx.send(f"Processing members for grandfathering Trusted...")
+
+        members_to_process = list(guild.members) # Create a list to avoid issues if members leave/join during processing
+        total_members = len(members_to_process)
+
+        for i, member in enumerate(members_to_process):
+            processed_count += 1
+            # Update status periodically
+            if i % 100 == 0 and i > 0:
+                await status_message.edit(content=f"Processing members... ({i}/{total_members})")
```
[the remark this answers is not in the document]
farmersrice: but at this point in time it has processed a total of i users, not i + 1, so it is ok as-is (for instance on i = 0 it has completed a total of 0 members out of total_members)

tle/cogs/handles.py
farmersrice: No, this should be at the bottom, because it is the least likely to be hit as no one is trusted. I did move the Purgatory check up though as that is most likely to be hit

tle/cogs/handles.py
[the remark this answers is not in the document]
algmyr: I read it as how many members has been processed, but whatever

tle/cogs/handles.py
```diff
@@ -830,6 +867,146 @@ async def role(self, ctx, action: str, which: str):
...
+        if trusted_role is None:
+            raise HandleCogError(f"The role '{trusted_role_name}' does not exist in this server.")
+
+        purgatory_role = discord.utils.get(guild.roles, name=purgatory_role_name)
+        # If Purgatory role doesn't exist, we assume no one has it.
+        if purgatory_role is None:
+            self.logger.warning(f"Role '{purgatory_role_name}' not found in guild {guild.name} ({guild.id}). Proceeding without Purgatory check.")
+
+        # The date when this code was added.
+        # April 20 was o3's first contest.
```
algmyr, line 938: Which one of these?

algmyr (approved):

PR #531, opened 2025-04-20
Adds support for custom emoji board

MKLOL:
Testing done

1. Migrated from an old server to confirm starboard messages remain and the starboard still works
2. Ran the server after the migration to confirm it doesn't migrate again
3. Added a pill react, with custom threshold
4. Confirmed it works and posts to the right channel, with the right color
5. Edited color and threshold and confirmed it works after

Chatgpt-ed most of it but PR is pretty straightforward

image

image

### DIFF
--- a/tle/cogs/starboard.py
+++ b/tle/cogs/starboard.py
@@ -8,14 +8,8 @@ from tle import constants
 from tle.util import codeforces_common as cf_common
 from tle.util import discord_common
 
-_STAR = '\N{WHITE MEDIUM STAR}'
-_STAR_ORANGE = 0xffaa10
-_STAR_THRESHOLD = 5
-
-
 class StarboardCogError(commands.CommandError):
     pass
-
 
 class Starboard(commands.Cog):
     def __init__(self, bot):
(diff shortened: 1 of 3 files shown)


MKLOL:
cc @cheran-senthil @algmyr

tle/cogs/starboard.py
```diff
@@ -74,75 +64,107 @@ def prepare_embed(message):
...
+            await ctx.send(embed=discord_common.embed_alert('Not found'))
 
     @discord_common.send_error_if(StarboardCogError)
     async def cog_command_error(self, ctx, error):
         pass
 
-
 def setup(bot):
-    bot.add_cog(Starboard(bot))
+    bot.add_cog(Starboard(bot))
```
cheran-senthil, line 170: Why remove the newline here?

Also, what is the point of color?

Do we want to keep calling this starboard or rename refs?

tle/util/db/user_db_conn.py
```diff
@@ -881,4 +972,4 @@ def remove_last_ratedvc_participation(self, user_id: str):
             return self.conn.execute(query, (user_id, vc_id)).rowcount
 
     def close(self):
-        self.conn.close()
+        self.conn.close()
```
cheran-senthil, line 975: Why remove newline here?

cheran-senthil (approved):
Meh, looks good to me @algmyr may have some comments

tle/cogs/starboard.py
MKLOL: cuz chatgpt is dumb i'll add it back.

Color is that slim green line to the left, here's for pillboard when i put red:

image

And default starboard:

image

cheran-senthil: Can you make the color be inferred automatically? I think fewer params are better

MKLOL: It has default yes. You don't need to specify

PR #535, opened 2025-05-24
Fix line length violations and set length to a proper value

algmyr:


cheran-senthil (approved):
Lint is passing, :bless:

Much thanks for taking the time to do this.

PR #536, opened 2025-06-16
Update user_db_conn.py

huythedev:
This PR fixes an sqlite3.OperationalError: no such column: befor in the TLE bot by correcting a typo in tle/util/db/user_db_conn.py.


cheran-senthil (approved):
Crazy that this was never caught till now lmao

PR #537, opened 2025-07-09
Stop using requirements.txt

algmyr:
It's outdated, just use pyproject.toml.


algmyr:
I tried this locally and the image built fine, let's try it and revert/fix it if needed.

